PR #1, opened 2020-03-28
Fix hunk #1 of League of Legends patch.

FLGX06:


Tk-Glitch:
Nice catch! Thank you <3

PR #3, opened 2020-03-29
Bail on a non-existent passed external config path, instead of blindly forging ahead without the user's awareness

Valmar33:
When an external config path is passed to proton-tkg.sh or non-makepkg-build.sh, and it doesn't exist, we should bail early, so that the user knows that they screwed up the path.

We don't want them puzzling over what went wrong during building, or with the build, if it completed, so alert them immediately.


Tk-Glitch:
Thanks!

PR #4, opened 2020-03-29
Add patch names to the md5 sums.

FLGX06:
Generated by a node.js script, seems to be correct.


openglfreak:
Does it stay when you do `updpkgsums`?
If not, that's impractical.

FLGX06:
I didn't know that's a thing. Yeah, it doesn't stay. If I knew that command exists I probably wouldn't make this PR.

PR #9, opened 2020-04-03
Fix League of Legends patch. (again)

FLGX06:


Tk-Glitch:
Here it comes :) Thanks!

PR #22, opened 2020-04-17
Fix integrity check error

FLGX06:
> ==> ERROR: Integrity checks (md5) differ in size from the source array.

Added missing checksum which wasn't added in commit 08b9d8cc562429d64c89beb8d9afd643a684899d.


Tk-Glitch:
Indeed. Thanks.

PR #31, opened 2020-05-03
proton-tkg: Add missing dxvk config variables to advanced config

FLGX06:


Tk-Glitch:
Those are in the main `proton-tkg.cfg` https://github.com/Frogging-Family/wine-tkg-git/blob/master/proton-tkg/proton-tkg.cfg#L53
We *could* move them to advanced considering they aren't really useful as basic config switches, I guess, but it's in one or the other. It can't be in both.

PR #52, opened 2020-05-24
Implement _dxvk_version

junaru:
Stumbled upon _dxvk_version in config files, didn't look like it was implemented yet, so i did.

* Must be **release tag**, not git.
* Defaults to "latest"
* Prompts on malformed value


Tk-Glitch:
Looks fine to me. Thank you!

PR #55, opened 2020-05-29
Change from True to Mainline in VKD3D

iWeaker4you:
When doing a clean installation on Arch Linux, it does not do the installation of VKD3D, see that in the PKGBUILD file was the selection of TRUE when it is already "mainline", but I do not know how the Fork installation will take


Tk-Glitch:
I'll make a follow-up change to that, but yes you're right. Thanks!

PR #82, opened 2020-07-11
wine-tkg: Debian packaging on Arch

FLGX06:
This allows wine-tkg to create debian packages on arch
Can be enabled with _GENERATE_DEBIAN_PACKAGE in config

TODO: Support different prefixes than /usr
TODO FIX: Make wine64-tkg the default as "wine64", the problem is it causes wine to complain about wine64 (invalid ELF header I think)

I didn't fully test this, but winecfg and wine explorer.exe works.
Hope this will be useful for people using Debian-based distros

### DIFF
--- a/wine-tkg-git/wine-tkg-scripts/package-debian.sh
+++ b/wine-tkg-git/wine-tkg-scripts/package-debian.sh
@@ -1,0 +1,101 @@
+#!/bin/sh
+
+files=$1
+prefix=$2
+output=$3
+name=$4
+pkgver=$5
+pkgname=$6
+tmpDir=/tmp/debian-package
+
+# TODO: Support different prefixes
+if [ "$prefix" != "/usr" ]; then
+    echo "Sorry, $prefix as prefix is not supported yet in wine-tkg packaging for Debian on Arch"
+    exit
+fi
+
+# Filesystem that can be found in $files
+lib32name="lib32"
+lib64name="lib"
+
+rm -rf $tmpDir || true
+mkdir $tmpDir
+
+# Package format version number
+echo "2.0" > $tmpDir/debian-binary
+
+# TODO FIX: Make wine64-tkg the default as "wine64", the problem is it causes wine to complain about wine64
+# Make wine-tkg and wine64-tkg the default
+mv "$files"/usr/bin/wine "$files"/usr/bin/_wine
+#mv "$files"/usr/bin/wine64 "$files"/usr/bin/_wine64
+cp "$files"/usr/bin/wine-tkg "$files"/usr/bin/wine
+#cp "$files"/usr/bin/wine64-tkg "$files"/usr/bin/wine64
+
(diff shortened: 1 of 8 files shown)


wine-tkg-git/wine-tkg-scripts/package-debian.sh
```diff
@@ -0,0 +1,101 @@
...
+Architecture: all
+Maintainer: Tk-Glitch <ti3nou@gmail.com>
+Homepage: https://github.com/Frogging-Family/wine-tkg-git
+Provides: ${pkgname}, wine, wine-development, wine-staging, libwine, fonts-wine, wine32, wine64
+Conflicts: wine, libwine, fonts-wine, wine32, wine64
+Replaces: wine, wine-development, wine-staging, libwine, fonts-wine, wine32, wine64
+License: LGPL
+Depends: libc6 (>= 2.17), libfontconfig1 (>= 2.11), libfreetype6 (>= 2.2.1), libncurses5 (>= 6), libasound2 (>= 1.0.16), libgcc1 (>= 1:3.0), libglu1-mesa | libglu1, liblcms2-2 (>= 2.2+git20110628), libldap-2.4-2 (>= 2.4.7), libmpg123-0 (>= 1.6.2), libopenal1 (>= 1.14), libpcap0.8 (>= 0.9.8), libpulse0 (>= 0.99.1), libx11-6, libxext6, libxml2 (>= 2.9.0), ocl-icd-libopencl1 | libopencl1, ocl-icd-libopencl1 (>= 1.0) | libopencl-1.1-1, zlib1g (>= 1:1.1.4)
+Installed-Size: $(du -sb ${files} | awk '{printf "%1.0f\n",$1/1024}')
+Description: 🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸
```
Tk-Glitch, line 87: The description found in the PKGBUILD could be used here as well with minor edits - to be a bit less sketchy. Maybe with a few frogs still :frog:

Tk-Glitch:
Also about that wine64 situation, what is it complaining about? I would have expected the WINEDLLPATH export to also affect `wine64` getting triggered by running `wine`.

wine-tkg-git/wine-tkg-scripts/package-debian.sh
FLGX06: Changed, I was happy that it works and forgot to change it :frog:

FLGX06:
So yeah, this should be ready as discussed on Discord without the `wine64` wrapper being the default.

Tk-Glitch:
Also enforcing it to disabled for proton-tkg will be required unless you have plans to add proton-tkg support later on. Disabling it in the meantime seems reasonable still.

FLGX06:
Done

Tk-Glitch:
Thanks! Let's go.

PR #102, opened 2020-08-07
Don't disable winevulkan_vkGetPhysicalDeviceSurfaceCapabilitiesKHR on staging 23ae4e6+

tannisroot:
The patch was removed


tannisroot:
Oh, actually I think this needs an additional check to make sure it is reverted when you bring back esync, let me fix this.

PR #103, opened 2020-08-08
Fix version tags not getting applied.

openglfreak:
Fixes output of `wine --version` not showing version tags, which in turn fixes various checks in Lutris.


Tk-Glitch:
:ok_hand: Thanks

PR #105, opened 2020-08-13
proton-tkg: DXVK artifact build support (from git.froggi.es)

FLGX06:
So this adds `$_use_dxvk="latest"` option which downloads latest job artifact of master branch from git.froggi.es as a zip file (I can't find tar file there).

Might be a good idea to consider having it as the default option rather than release being the default as I think it would fit proton-tkg as it's already using development wine, staging for example (at least by default).


Tk-Glitch:
froggi.es is down far too often to be default IMHO, but having the option available sounds good.
Thanks :+1:

PR #106, opened 2020-08-13
Allow using upper-cased "Y" to answer questions

FLGX06:
Note: I think this statement is true if the answer contains "y" or "Y"


Tk-Glitch:
I'm okay with that, thanks!

PR #118, opened 2020-08-27
Add package name check to replaces declaration

kode54:
Fixes cases like 'protonified' preset, which produce a package
named 'wine-tkg-git-protonified', and then have a replaces that
points at the same package name, which does funny things when
you add builds to a custom repo, like constantly asking you to
re-download and re-install the package.


Tk-Glitch:
Well spotted! Thanks :frog: :heart:

PR #164, opened 2020-10-04
Proton-tkg: Add vulkan-headers as make dependency

FLGX06:
Looks like it's needed for vrclient build otherwise vrclient build fails on `<vulkan/vulkan.h>` include (and then steam fails to launch games with proton-tkg)


Tk-Glitch:
Thanks!

PR #175, opened 2020-10-10
deps: Don't use sudo when installing packages as root

imaami:
In Debian sudo is an optional package and might not always be installed,
one example being the official Debian Docker images. A wine-tkg "debuntu"
build in a minimal Debian container will fail even when run as root due
to `_debuntu_{32,64}()` always using sudo.

Change the package installation commands' hard-coded "sudo" prefix to a
variable whose value is either "sudo " or and empty string depending on
the effective user ID.

Signed-off-by: Juuso Alasuutari <juuso.alasuutari@gmail.com>


Tk-Glitch:
Thank you! I guess that'll please many.

PR #259, opened 2021-01-08
Add _staging_userargs option to proton-tkg advanced-customization.cfg

Valmar33:


Tk-Glitch:
That was coming :frog: 
Could you please add it to the sample config as well?

Valmar33:
Done. :frog:

PR #260, opened 2021-01-10
proton-tkg: Remove "latest" DXVK option which downloads from git.froggi.es

flgx16:
git.froggi.es was changed from GitLab to Gitea and no longer has DXVK artifacts


Tk-Glitch:
Thanks :heart:

PR #276, opened 2021-01-28
Fix conhost.exe taking 100% cpu with fsync.

openglfreak:


openglfreak:
:thinking: Now Overwatch doesn't start, neither with esync nor with fsync. I'll try without this change

openglfreak:
Ok, good news is it's not this change's fault, bad news is it also doesn't start without this change...

Tk-Glitch:
The conhost issue is gone on my end so far. Has been for weeks.

openglfreak:
It's not gone for me. Look at the changes in this PR, they make sense, right?

Tk-Glitch:
Absolutely

Issue #248, opened 2020-12-27
Gamepad not working on any 6.0 builds


JooJooBee666:
I can no longer see my gamepad on any of the 6.0 builds from the control panel and games don't recognise it either, same for both wired 360 controller and wireless Xbox one controller.  I can only see a disabled (js) device (see image).

![gamepad](https://user-images.githubusercontent.com/1514297/103163073-f554bc80-47c6-11eb-9900-fafeb21e6732.jpg)

I also tried the build you posted here (proton_tkg_6.0rc3.r0.g9912133b.release) but it has the same issue as well.  I recompiled with the "_gamepad_additions="true" flag, but this made no difference.  I can attach the log from the builds but seeing as the official release here is also broken, I see no point.  Maybe it's an issue with SDL?  Not really sure where to go from here.  Please let me know what else you need to troubleshoot this.

System:
Ubuntu 20.10
Kernel 5.8.0-33-generic

Valmar33:
Have you tried clicking on the disabled entry, and seeing if it becomes available to enable?

Just tried on my end ~ it defaults to not offering to enable, until I click on the entry proper, and then it gives me the choice to.

JooJooBee666:
I should have mentioned that.  Yes, but when I do it just disappears, it doesn't show up in the connected box, it's just GONE.  If I restart the control panel, it's back to disabled.

JooJooBee666:
Just to further clarify, the one that works correctly is just named "Xbox One Wireless Controller", which is missing now, no (js) appended.

sonic2kk:
Potentially related: #239, #231

JooJooBee666:
> Potentially related: #239, #231

Agreed.  I saw those too but I did a bit more digging and found it was for anything gamepad related.

Where upstream is this bug mentioned? Anyone know?  I don't see any patches at https://source.winehq.org/patches/ that seem to mention anything about gamepads, dinput, xinput or joysticks.

Iglu47:
I saw https://bugs.winehq.org/show_bug.cgi?id=50321#c1 and share here, sorry if not related

logos128:
[...]

Issue #231, opened 2020-12-07
Yakuza: Like a Dragon (1235140) does not recognise controllers


sonic2kk:
As of 6.0-RC-1, *Yakuza: Like a Dragon* does not recognise controllers. In Steam, it does not even show the controller configuration being loaded. When trying to enter inputs, nothing happens. Button prompts do not change from keyboard/mouse to controller, it is as if there is no controller connected at all. Proton-tkg 5.22 and below recognised controllers, and indeed Proton 5.13 recognises controllers. Proton-GE also works.

I tested some other games and they seem to recognise controllers fine with 6.0-RC-1, so far I haven't found any other games affected by this issue. The other Yakuza games, namely _Yakuza Kiwami 2_ which uses a similar engine, do not have any controller issues.

**Note**: Controllers for this game are incorrectly mapped due to a general game bug, so this isn't _too_ big of an issue as the game is unplayable with a controller at this moment in time. The point of this issue is to report that the controller isn't recognised at all.

Tk-Glitch:
It *might* need `_gamepad_additions="true"` (which is not enabled by default in -tkg due to some other issues with different games). That patchset is part of both upstream proton and GE builds.

sonic2kk:
Perhaps I enabled it wrong, but that didn't seem to work. I set the flag in `proton-tkg/proton-tkg.cfg`. And for clarity's sake for anyone reading this in the future, this source build was built against Wine 6.0-rc2 as opposed to rc1.

sonic2kk:
Just updating to say the issue is still present in 6.0-rc3.

sonic2kk:
To update: As of today's latest Proton Experimental build, controllers in *Yakuza: Like a Dragon* function correctly.

Tk-Glitch:
Apparently it breaks steering wheels in other games as a result. The fun never ends :frog: 
Importing the change for yakuza behind a toggle sounds like the best idea.

HurricanePootis:
> To update: As of today's latest Proton Experimental build, controllers in _Yakuza: Like a Dragon_ function correctly.

Did you build it with the game pad additions enabled, or not? I am having issues with controller detection in 6.0rc6 (nothing changed in build configs) on Yakuza 0.

Tk-Glitch:
It might not be fixed (I don't have the game to check), so I'll reopen if the patch wasn't enough. Please let me know how it goes.

sonic2kk:
[...]

PR #279, opened 2021-01-30
Update proton-sdl-joy patchset for e4fbae8 and move previous version to legacy.

logos128:
After looking further at the issue with that commit https://github.com/wine-mirror/wine/commit/e4fbae832c868e9fcf5a91c58255fe3f4ea1cb30 , found that the problem is with the obsolete "HAVE_SDL2_SDL_H" define in the proton-sdl-joy patchset. Since it's no longer defined by the commit, it causes those file parts to be excluded from compilation, hence no controllers in wine.

So there is no problem with the upstream commit:) Libinput controllers work perfectly with it, as well as SDL2 controllers in proton-tkg with the proposed changes in this PR.

Properly fixes https://github.com/Frogging-Family/wine-tkg-git/issues/248 and https://github.com/Frogging-Family/wine-tkg-git/issues/231, etc.


Tk-Glitch:
Great! Thanks for diving in! :frog: :heart:

PR #289, opened 2021-02-16
Fix Proton-TKG compiling

iWeaker4you:
I'm new to Wine, but what little I know about programming, gave me to understand that the "return" was wrong XD


Tk-Glitch:
Thanks! Nice typo from me indeed :frog:

PR #300, opened 2021-03-19
Update FFXIV patch so it works again

Valmar33:
FFXIV patch was broken due to the extensive changes that have happened with ntdll over the past many months.

This small fix should make that go away. Until it breaks again...

I don't know what commit the current FFXIV patch successfully applies to, and I don't know how to figure that out, so can I leave that to you? frog_donut


Tk-Glitch:
Sure. Thanks!

PR #345, opened 2021-04-27
Fix typo 🐸

IroAlexis:
There are spaces that got lost. It is a little bit stressful :sweat_smile:


Tk-Glitch:
Nice catch! Thanks :frog: :heart:

PR #417, opened 2021-06-16
Fix the commit id for the 32-bit hotfix

tannisroot:
🐸


Tk-Glitch:
Thanks tannis! :frog: :heart:

PR #437, opened 2021-06-30
profiles: Move chaotic-lol to -opt; Drop chaotic-wayland

PedroHLC:
#### :thinking: What?

1 - Move chaotic-lol to `-opt`;
2 - Drop chaotic-wayland (for now);

#### :man_shrugging: Why?

About (1), so users don't mix their wine setups for a single game;
About (2), it needs reverting and tweaking, let's just push it when it's ready/staging;

#### :clipboard: Pending

Nothing.


Tk-Glitch:
Outside of the needed env vars, we should handle all that's needed since https://github.com/Frogging-Family/wine-tkg-git/commit/3e594de49f56944dae138969d59d85a6f7ec9f0b
But yeah it doesn't offer anything actually worth using for end users right now.

Issue #237, opened 2020-12-13
wine-tkg-git: Issues with makechrootpkg, jack/jack2


krathalan:
This issue is for wine-tkg-git, not proton-tkg.

If I try to use `makechrootpkg` with a clean chroot, I get the following error when installing dependencies before the build:

```
==> Checking buildtime dependencies...
==> Installing missing dependencies...
resolving dependencies...
looking for conflicting packages...
:: jack2 and jack are in conflict. Remove jack? [y/N]
error: unresolvable package conflicts detected
error: failed to prepare transaction (conflicting dependencies)
:: jack2 and jack are in conflict
==> ERROR: 'pacman' failed to install missing dependencies.
```

Specifying `jack2` in the `depends=()` array in the PKGBUILD fixes this issue and allows me to build wine-tkg-git.

Thaodan:
Whats missing to fix this? This also affects plain Arch Linux, I'm not sure how this is distribution related.

Tk-Glitch:
The issue is due to cross-dependency. Jack or Jack2 are no explicit dependency for wine. It seems to be a conflict somewhere possibly with the `ffmpeg` package which depends on the `jack` package explicitly, followed by another package which depends on `jack2` explicitly. I am not sure yet what depends on jack2 explicitly but that would need to be fixed there instead of on our side.

ViNi-Arco:
Good afternoon, people..

The reason for that is, with **lib32-gst-plugins-good 1.18.1**, the dependency on [lib32-jack](https://github.com/archlinux/svntogit-community/commit/874c77eff0e29e7af589b8c4da8fa42b2f8db52a#diff-37538beb61ff63edebbf735dfcf39e5d732f49183d6beb097169d971875ca422L12-R12) has been added.
And Etienne's script has [lib32-gst-plugins-good](https://github.com/Frogging-Family/wine-tkg-git/blob/master/wine-tkg-git/PKGBUILD#L114) as a build dependency, with this he informs the replacement of **lib32-jack** with **lib32-jack2**.

this during the build. then you can remove **lib32-jack2** and install **lib32-jack** again if you want it.

PR #447, opened 2021-07-09
closes #243, closes #443; add initial Github Actions CI support

### LINKED ISSUE #243: build automation with CI?
Github Actions & ~~Gitlab CI~~ provides a huge free credit theoretically enough for several distributions. Any plans to implement this?

### LINKED ISSUE #443: Automatic CI using wine-tkg-git
So some members of the grapejuice community (A Roblox on Linux community) including I are seeking to automatically compile wine with patches specific to the Roblox client. And we like using wine-tkg-git to do this since it facilitates the whole process really well, so I wanted to ask if it's possible to supply automatic input to the non-makepkg-build.sh script so we can automate the compile process or should I be doing something else?

paulcarroty:


.github/workflows/wine-arch.yml
```diff
@@ -0,0 +1,29 @@
+name: Wine Archlinux CI
```
CodingKoopa, line 1:
```suggestion
name: Wine Arch Linux CI
```
See [here](https://wiki.archlinux.org/title/Arch_terminology#Arch_Linux).

.github/workflows/wine-arch.yml
```diff
@@ -0,0 +1,29 @@
...
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: archlinux:latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Compilation
```
CodingKoopa, line 17: Names have inconsistent tenses - perhaps change this to `Compile`

.github/workflows/wine-arch.yml
```diff
@@ -0,0 +1,29 @@
...
+        run: |
+          echo -e "[multilib]\nInclude = /etc/pacman.d/mirrorlist" >> /etc/pacman.conf
+          yes|pacman -Syu --noconfirm base-devel sudo
```
CodingKoopa, line 20: Is this `yes` necessary? I would expect `--noconfirm` to take care of that.

CodingKoopa, line 20: Does this actually work, without running `pacman-key` as described [here](https://hub.docker.com/_/archlinux)?

CodingKoopa, line 20: Are you sure `sudo` is necessary, even if we use `su` later on?

.github/workflows/wine-arch.yml
```diff
@@ -0,0 +1,29 @@
...
+          useradd user -G wheel && echo "user ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
```
CodingKoopa, line 21: (ditto, about whether `/etc/sudoers` is necessary)

.github/workflows/wine-arch.yml
```diff
@@ -0,0 +1,29 @@
...
+          chown user -R . && cd wine-tkg-git
+          su user -c "yes|PKGDEST=/tmp/wine-tkg makepkg --noconfirm -si"
```
CodingKoopa, line 23: Unnecessary `yes`?

CodingKoopa, line 23: Unnecessary `--install`?

CodingKoopa (changes requested):
Thanks! I added some initial notes on the Arch Linux CI.

.github/workflows/wine-arch.yml
paulcarroty: `--noconfirm` failed several times, so added `yes` to be sure.

paulcarroty: It's working, I can add the key if it really necessary.

paulcarroty: Build scripts required it.

.github/workflows/wine-arch.yml
paulcarroty: Non-root user required.

.github/workflows/wine-arch.yml
paulcarroty: Doesn't works without `yes`.

paulcarroty: Fixed.

.github/workflows/wine-arch.yml
CodingKoopa: Okay, as long as it works that should be okay.

CodingKoopa: Ah, I didn't think about that ^^

.github/workflows/wine-arch.yml
paulcarroty: Deleted all `yes` in Arch CI.

CodingKoopa: I thought about it some more; this one is actually probably needed. I forgot that the `PKGBUILD` is interactive; you probably will need this particular one in order to select the default values for the TkG configuration.

paulcarroty:
@CodingKoopa seems like all resolved, but got #237 issue with jack&jack2 on Arch.

.github/workflows/wine-arch.yml
paulcarroty: Done.

.github/workflows/wine-ubuntu.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Compilation
+        run: |
+          sudo dpkg --add-architecture i386 && sudo apt update
+          cd wine-tkg-git 
+          sed -i 's/distro=""/distro="debuntu"/' customization.cfg
+          yes|./non-makepkg-build.sh
+      - name: Archive the artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: wine-tkg-build
```
AdelKS, line 26: You are capturing the wrong folder, the correct one is `non-makepkg-builds`. For now the artifacts are 3.4Gb x)

paulcarroty: Fixed.

paulcarroty: Your suggestion points to empty dir 😄

paulcarroty: Fixed.

Tk-Glitch:
Looks fine to me for its current purpose. There was a kinda unified route pending but I guess it could be simplified now with some additional work on linux-tkg prior. Thanks! Merging.
:frog: :heart:

robigan:
Uhh big issue, I get a `fatal: could not create work tree dir 'community-patches': Permission denied` error when using community-patches flag

PR #448, opened 2021-07-10
Update proton-sdl-joy patchset for a173672 and 2bd3c97, and move prev…

logos128:
-Implements all the changes made by upstream after wine-mirror/wine@a173672, including the substantial refactoring of dinput and hidclass.sys between wine-mirror/wine@5fe1031 and wine-mirror/wine@2bd3c9703.

-Since the changes are too frequent, the patchset is disabled between wine-mirror/wine@5fe1031 and wine-mirror/wine@2bd3c9703 (around v6.10).

-Includes legacy version with the changes made after wine-mirror/wine@a173672 till wine-mirror/wine@5fe1031 (v6.7, v6.8 and v6.9).

-The patchset is compatible with mainline since at least wine-mirror/wine@a173672, so it's also supported.

-Enables the patchset for wine-tkg (turned off by default), due to the important fixes it includes for FFB steering wheels, as well as the proper support for HIDRAW access.

-The patchset doesn't depend on any of the other proton patches, so it should be easier to maintain.

I'm using it for a few weeks now without any problems with my FFB steering wheel and pedals. Works well with all simracing games in both wine-tkg and proton-tkg (both staging and mainline) :)


Tk-Glitch:
That's pretty cool! It looks fine at first glance.
I'll test for the known bug we had before - which lead to the patchset getting disabled initially - on a couple games + some specific gamepads. If it fixes FFB steering wheels it's worth having the option nevertheless.
Thank you! :frog: :heart:

PR #469, opened 2021-07-29
Improve comments and skip some patches in the earlyhotfixer script.

satyrnsstuff:
Skips some patches if they've already been patched in the Wine version selected.

The point of this is to avoid patching errors and make the patching process smoother.


Tk-Glitch:
The Star Citizen community patch isn't the same as 7d7b8e2a88c58e05fb80d3600a52c473ddee4c5d (and has nothing to do with it at this point). It's not required anymore to play the game (as the dependency on added function has been removed from the game since), but the function is still currently unimplemented upstream so the patch still applies, and might be needed for other games/apps using `StorageDeviceSeekPenaltyProperty`.

I'm going to nuke that patch for now.

Tk-Glitch:
Thanks! :frog:

PR #472, opened 2021-07-31
Skip another patch.

satyrnsstuff:
Skips the Add-SORT_DIGITSAS-UMBERS-flag-to-CompareStringsEx patch if the commit that fixes it is in the Wine version selected.

The upstream commit and the patch are not the same, but the commit still adds everything the patch does, plus some extra stuff.


Tk-Glitch:
Thanks! :frog: :heart:

PR #533, opened 2021-10-04
Hotfixer: Remove a TRACE in the DXGIGetDebugInterface1 stub causing segfaults.

Fox2Code:
This error was triggered by the unity crash reporter.

I tried and tested the patch before posting the pull request, and I can say the error reporter is no longer crashing...  
I can't say the same thing for the game.

I know it's not a lot but I hope it helps.
Also since I saw you liked to support old wine versions, I tried to make it support old versions too.


Tk-Glitch:
I'm fine with that :) Thank you! :frog:

Issue #519, opened 2021-09-23
Feature request:Profile-Guided Optimization (PGO) 


123-123-nil:
It would be great if there were an option like `ENABLE_PGO` enabling that should compile wine-tkg with `-fprofile-generate` then generate a profile  by running various workloads like `make test` and other user specified applications, after that recompile wine with the generated profile. I am not a developer. Just an end user of your awesome builds. Got this idea by reading [this](https://documentation.suse.com/sbp/all/html/SBP-GCC-10/index.html)

PR #536, opened 2021-10-07
Partially enable linker time optimization.

Fox2Code:
Partial because when added to `_CROSS_LD_FLAGS` it crash the build, but on `_LD_FLAGS` it work fine.

I found out about LTO because in the document @123-123-nil attached in the #519 issue about PGO.

### DIFF
--- a/proton-tkg/proton-tkg-profiles/advanced-customization.cfg
+++ b/proton-tkg/proton-tkg-profiles/advanced-customization.cfg
@@ -70,7 +70,7 @@ CUSTOM_GCC_PATH=""
 
 _LOCAL_OPTIMIZED="true"
 _GCC_FLAGS="-O2 -ftree-vectorize"
-_LD_FLAGS="-Wl,-O1,--sort-common,--as-needed"
+_LD_FLAGS="-Wl,-O1,--sort-common,--as-needed -flto"
 _CROSS_FLAGS="-O2 -ftree-vectorize"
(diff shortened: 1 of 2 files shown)


Fox2Code:
Forgot to add the arg to `proton-tkg` the first time, so I force pushed to avoid making too many commits.

Fox2Code:
Done, the new flag should be clang compatible too now.

Tk-Glitch:
I'll never merge that, just so it's clear from the start. Adding an option for it sounds too far fetched also. One can set their own compiler flags as desired already. Feel free to keep this open in case some people wanted to add the flag for their builds, and you can use this PR to talk about/experiment with compiler flags if you want.

Fox2Code:
Ok I completely understand, it's your wine, so it's your choice. 
As long as you don't prevent other to add this to their custom wine builds it's ok for me.
Sadly I will had to close this PR at some point because it target the wine branch I use to contribute to this repo.

Also is there a community place like any kind of Forum of Discord to talk about this?
I searched anywhere but didn't find any kind of place outside GitHub to talk about about wine-tkg.

Tk-Glitch:
Yes we have a Discord server - https://discord.gg/jRy3Nxk
Even with the PR closed we could link to it from a meta-issue or another PR, I guess that would be fine.

Fox2Code:
Thanks for the Discord link.

Fox2Code:
Closing, everyone can still add the parameter in their config files.

PR #538, opened 2021-10-07
Enable building with Clang

AdelKS:
I toyed around building wine with Clang,

Note that wine has two parts: native and cross-compiled for windows, and the compiler can be arbitrarily chosen for each part. This change enables choosing which compiler for which part. The wine build system supports clang when it detects that the user asked for it.

It indeed produces a functional wine when building wine-mainline without any extra patch, otherwise staging produces errors. And, if using all wine-tkg patches, the `steam.patch` makes an error with `lld-link: error: undefined symbol: _StrStrW`. I do not know if there would be more if that one gets fixed.

Another issue, the produced `wine` executable is 2GB, I guess the stripping tool needs to be updated or something.

I did not plan to open this WIP PR, but people can at least see it and participate.

Adel

### DIFF
--- a/proton-tkg/proton-tkg-profiles/advanced-customization.cfg
+++ b/proton-tkg/proton-tkg-profiles/advanced-customization.cfg
@@ -77,6 +77,14 @@ _NUKR="true"
 _NOCOMPILE="false"
 _NOINITIALPROMPT="false"
 
+# Replace with "clang" and "clang++" to use LLVM Clang
+_C_COMPILER="gcc"
+_CXX_COMPILER="g++"
+
+# Replace with "clang" (for both) to use LLVM Clang
+_CROSS_C_COMPILER_64="x86_64-w64-mingw32-gcc"
+_CROSS_C_COMPILER_32="i686-w64-mingw32-gcc"
+
 # Optionally set additional make dependencies for makepkg builds. Multiple elements should be separated by a space.
 # Only affect makepkg
 _user_makedeps=""
--- a/wine-tkg-git/wine-tkg-profiles/advanced-customization.cfg
+++ b/wine-tkg-git/wine-tkg-profiles/advanced-customization.cfg
@@ -53,6 +53,14 @@ _CROSS_FLAGS="-O2 -ftree-vectorize"
 # Same as _LD_FLAGS but for cross-compiled binaries.
 _CROSS_LD_FLAGS="-Wl,-O1,--sort-common,--as-needed"
 
+# Replace with "clang" and "clang++" to use LLVM Clang
+_C_COMPILER="gcc"
+_CXX_COMPILER="g++"
+
+# Replace with "clang" (for both) to use LLVM Clang
+_CROSS_C_COMPILER_64="x86_64-w64-mingw32-gcc"
+_CROSS_C_COMPILER_32="i686-w64-mingw32-gcc"
+
 # By default, tests are disabled to speed up compilation. If you need them for development purposes, set to "true"
 _ENABLE_TESTS="false"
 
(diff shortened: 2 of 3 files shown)


Tk-Glitch:
Pretty much the same thing as the other PR. It's very unlikely I'll ever merge this (even less so than the other, actually), but it's fine to have the PR open for the ones who want to experiment.

Tk-Glitch:
I'm simply not convinced of its usefulness (see my message above). Wine is bound to have a hard dependency to mingw in the near future, making clang pretty problematic for pretty much the whole userbase but a couple users.

Seeing how everyone is jumping on the clang option in linux-tkg then complain about their nvidia modules not working, I'm kinda frightened about merging this. Wine already breaks in various ways when built with plain gcc without mingw. This adds fuel to the fire.

I'm thinking about making it a hidden option as a compromise, defaulting to our current values but allowing bypassing them from .cfg, but not explicitly offered there.

paulcarroty:
Can it be added to a custom script, something like `experimental-clang-build.sh`? GCC/LLVM flags aren't 100% identical.

One source and two compiler builds - great opportunity for bug catching.

AdelKS:
> Can it be added to a custom script, something like experimental-clang-build.sh? GCC/LLVM flags aren't 100% identical.

I had to touch some underlying files too, so the changes cannot be contained in a single extra file.

> One source and two compiler builds - great opportunity for bug catching.

Tell that to upstream wine to setup CI with GCC & Clang builds. I wonder if they do actually, probably not :thinking:

pongo1231:
> Another issue, the produced wine executable is 2GB, I guess the stripping tool needs to be updated or something.

Have you tried to add the `-std=gnu89` flag to your compile flags as mentioned in [here](https://wiki.winehq.org/Clang)? That seems to have severely reduced the size of the binary to normal levels for me. I didn't need the `-g` flag however.

AdelKS:
I did find that page but it's rather old... I think no information there is useful anymore x) But it's worth trying

KuKuRuZ1337:
It turned out very well in the end.
The main thing is not to use x86_64-w64-mingw32-clang and i686-w64-mingw32-clang.
I'll use C_COMPILER=clang and CXX_COMPILER=clang++ every time now
Thanks

AdelKS:
> I'll use C_COMPILER=clang and CXX_COMPILER=clang++ every time now

Yea that works, cross-compiling with clang is the remaining issue.

Fox2Code:
cross-compiling use something called `winegcc` as a wrapper to `x86_64-w64-mingw32-gcc` or `i686-w64-mingw32-gcc`.

But I still find this PR useful for example compiling x86_64/amd64 wine on ARM, this PR is not just a clang enable one.

Anyway, my wine-tkg fork can use `sed` to patch wine-tkg scripts and use clang directly if I want to, so not merging this doesn't really prevent anything.

Also there is also still a way to fake gcc as clang via the configuration file by creating link to clang named gcc and adding the directory into the PATH variable.

This is why I feel it's a good idea to merge this, because peoples that want to use clang can already do this with hacky solutions, and it simplify something that is originally hard to accomplish.

Personally I don't really care about clang this patch add, I care about config being consistent, as the clang support can already be added by other means than this patch.

AdelKS:
> cross-compiling use something called winegcc as a wrapper to x86_64-w64-mingw32-gcc or i686-w64-mingw32-gcc.

An interesting thing to know is that the wine compile script will actually look for `x86_64-w64-mingw32-clang` _if it doesn't find first_ `x86_64-w64-mingw32-gcc`. But, I did get an `llvm-mingw` build [from here](https://github.com/mstorsjo/llvm-mingw), literally uninstalled mingw-gcc but the compilation fails while looking for `libgcc.so` for linking.

Fox2Code:
Well my point is this is more a quality of life patch than anything else, forcing wine-tkg to use clang as a compiler is already possible with hacky solutions, this is why I would like this to be merged, to avoid peoples from using hacky solutions.

paulcarroty:
Yeah, it's better to have the clang option.

KuKuRuZ1337:
can you update it so I can use clang to build it ?

PR #557, opened 2021-10-23
Make community-patches logic play nicely with makechrootpkg

rplociennik:
This patch attempts to make the logic of locating/creating `community-packages` repo root more reliable, especially in `makechrootpkg` environment.


Tk-Glitch:
Thanks! :frog: :heart:

PR #565, opened 2021-10-27
Add low latency pulse/pipewire patch stolen from ThePoon

X9VoiD:
## Changes

- Added `wine-tkg-git/wine-tkg-patches/misc/lowlatency_audio_pulse.patch`.
- Modified `wine-tkg-osu.cfg` profile to use `_lowlatency_audio_pulse=true`.
- Updated `wine-tkg-git/wine-tkg-scripts/prepare.sh` to support `_lowlatency_audio_pulse` configuration.
> The patch is applied only if [f77af3dd632](https://github.com/wine-mirror/wine/commit/f77af3dd6324fadaf153062d77b51f755f71faea) is applied to current HEAD, which is the earliest commit that the patch applies.

This patch works for both PulseAudio and Pipewire.


Tk-Glitch:
Thanks guys! :frog: :heart:

PR #568, opened 2021-10-28
split memset-fix

Iglu47:
This fixes is necessary for the correct applying of the of memset-fix on older versions where it was required.
And adds missing parts for proton-tkg after 1dbbc803ac5dc138e4362b6555a786296f2b1f0b


Tk-Glitch:
Thanks Iglu47! :frog: :heart:

PR #576, opened 2021-11-01
Improve fsync_futex_waitv patch compatibility

Iglu47:
Adds https://gist.github.com/ivyl/93107ffef32b2e8816d1176286afdd12

This will not make it possible to keep multiple interfaces (unchanged) in the same kernel build, it just adds safe futex_waitv disabling for kernels with old futex2 interface.


Tk-Glitch:
It's not a critical addition but it'll indeed help with the transition.

Iglu47:
patch was accepted as https://github.com/ValveSoftware/wine/commit/a55a995020e09bfb844b12cac02351cf970287c7

Tk-Glitch:
Thanks! :frog:

Issue #622, opened 2021-12-08
fastsync-staging patches fail to apply with latest wine-tkg (2cad614)


CartoonFan:
Hello there! I'm on Arch, and with 2cad614ba06f3fee2793a396b502a3b48a7cf5e9, the fastsync-staging patches don't apply correctly for me. Tried with a fresh clone as well, with no changes.

Hopefully these files will help:
[prepare.log](https://github.com/Frogging-Family/wine-tkg-git/files/7678588/prepare.log)
[last_build_config.log](https://github.com/Frogging-Family/wine-tkg-git/files/7678589/last_build_config.log)
[wine-tkg.cfg.txt](https://github.com/Frogging-Family/wine-tkg-git/files/7678594/wine-tkg.cfg.txt)
 
 Thanks as always!

Iglu47:
thanks for report, i'll see soon

CartoonFan:
Thanks!

Iglu47:
It's due to `_protonify="true"`. The same problem on proton-tkg.
This requires a little change, but I cannot successfully build proton-tkg right now due to some other error.

CartoonFan:
Thanks! I'll try it now.

PR #618, opened 2021-12-05
Update fastsync patches

Iglu47:
wine-side part of winesync / fastsync https://github.com/Frogging-Family/linux-tkg/pull/379
source: https://repo.or.cz/wine/zf.git/shortlog/refs/heads/fastsync3

Esync/Fsync compatible version of fastsync based on work @openglfreak: https://github.com/openglfreak/wine-tkg-userpatches/tree/cf05e1b6df08a9591948e28915c26521369ddbe5/patches/0001-fastsync

PR #623, opened 2021-12-08
Update fastsync patches for proton-tkg

Iglu47:
Required extension for https://github.com/Frogging-Family/wine-tkg-git/pull/618
Fixes build proton-tkg and wine-tkg with `_protonify="true"` option

Fixes https://github.com/Frogging-Family/wine-tkg-git/issues/622


Tk-Glitch:
Thanks! I see you also updated the cfg comment 🐸 ❤️

PR #626, opened 2021-12-09
Update and Improve fastsync compatibility with clock_monotonic

Iglu47:
Resolves conflicts using `_use_clock_monotonic` and `_use_fastsync`

And here I noticed that it is probably possible to skip some of the clock_monotonic patch checks since wine-tkg is set up to avoid using `ntdll-NtAlertThreadByThreadId`
(but let me know if it's unnecessary and you don't want it).

Update fastsync patches like https://github.com/Frogging-Family/wine-tkg-git/commit/853fb84bffa214478ca15bacb6939121267b000b
Fixes https://github.com/Frogging-Family/wine-tkg-git/issues/628


CartoonFan:
I pushed a patch to Iglu47:fastsync3-clock_monotonic to make the fastsync-staging-protonify patch work with 7.0-rc2, but I encountered a conflict afterwards with proton-cpu-topology-overrides that I don't know how to fix. The prepare.log is here if you're interested: [prepare.log](https://github.com/Frogging-Family/wine-tkg-git/files/7739912/prepare.log)

Iglu47:
> I pushed a patch to Iglu47:fastsync3-clock_monotonic to make the fastsync-staging-protonify patch work with 7.0-rc2, but I encountered a conflict afterwards with proton-cpu-topology-overrides that I don't know how to fix. The prepare.log is here if you're interested: [prepare.log](https://github.com/Frogging-Family/wine-tkg-git/files/7739912/prepare.log)

yes, this is the expected error and it's okay as this patch is applied to the same files as fastsync.
and until fastsync is enabled by default (it won't be soon) it would be necessary to maintain two versions of this patch (I suppose this will not happen).

CartoonFan:
Do you have an idea for a workaround? Commenting out the patch in prepare.sh causes a build error...

Iglu47:
> Do you have an idea for a workaround? Commenting out the cpu topology patch in prepare.sh causes a build error...

Sorry it wasn't clear.

Yes it doesn’t work properly (it looks like doing checks on wine-tkg is unexpectedly much more difficult than I expected).
Thanks for the report, I'll take a look tomorrow and add a fix also to this PR.

CartoonFan:
> > Do you have an idea for a workaround? Commenting out the cpu topology patch in prepare.sh causes a build error...
> 
> Sorry it wasn't clear.
> 
> Yes it doesn’t work properly (it looks like doing checks on wine-tkg is unexpectedly much more difficult than I expected). Thanks for the report, I'll take a look tomorrow and add a fix also to this PR.

All right, thanks. Sorry if I misunderstood you. I'll be waiting for the fix :grin:

Iglu47:
merged fastsync sections on `prepare.sh` and removes `fastsync-staging-prep.patch`

Iglu47:
actually doesn't make sense now because of https://github.com/Frogging-Family/wine-tkg-git/commit/d8b4bad308ea1bfe651097cb90ed2c393eba3bf8
although some simplifications will be used in the future (the clock_monotonic simplification would be especially useful)

Issue #643, opened 2021-12-24
Error building with futex_waitv patch


v1993:
The error I'm encountering is described here: https://github.com/ValveSoftware/wine/issues/131. It seems to be caused by having system headers that already define futex-related variable `__NR_futex_waitv`.

v1993:
Note: linked issue also provides a hotfix for this problem, which seems to work when added as user patch.

Iglu47:
what does it mean already defined?
are you talking about compatibility not only with linux-tkg? I have not tested with anything else.

v1993:
I'm using linux-tkg with futex_waitv patchset currently and that seems to be the exact source of problem.

Iglu47:
can't repro, can you share at least the .cfg files?

Iglu47:
@v1993 what version of this package do you have on your system?
https://archlinux.org/packages/core/any/linux-api-headers/
`pacman -Qs linux-api-headers`

v1993:
None, because I'm on KUbuntu :stuck_out_tongue:. I have version `5.15.11-tkg-pds-1` of `linux-libc-dev` tho (which was built with futex_waitv patch).

Iglu47:
Ok that makes sense.
Yes, it will be useful to explicitly separate those definitions by the number of files to which they are defined, but it will be necessary to add a check for include `linux/futex.h`. Looks simple, I'll take a look later.

Iglu47:
@v1993 can you download and check this patch for wine-tkg, please
https://github.com/Iglu47/wine-tkg-git/commit/17ea486c128a40fbb3882b221dd92d874601a6f1.patch
```
cd wine-tkg-git
patch -Np1 <  17ea486c128a40fbb3882b221dd92d874601a6f1.patch
```
I would like to know if this works correctly for another distribution before submitting it to TkG.

liberodark:
Hi,

Have see your patch i think is better to upstream your patch to https://github.com/ValveSoftware/wine/
If you are ok with that?

Best Regards

Iglu47:
well, over time the Proton build environment will use kernel-5.16+, and only then are such changes likely to be needed.

PR #646, opened 2021-12-26
futex_waitv: Add check for linux futex header

Iglu47:
I hope these small changes will be enough to solve the problem of building Wine-TkG in different build environments, and as I expect this should help solve the same problem with some future packages in Archlinux build environments.

Fixes https://github.com/Frogging-Family/wine-tkg-git/issues/643


Iglu47:
Just checking for `linux/futex.h` is not entirely appropriate in this case, since this file exists for kernels versions lower than 5.16 (but also they can be patched for futex_waitv). This PR does not add direct checks to the definition of FUTEX_32 and struct futex_waitv, but it should still be enough to get them from `linux/futex.h`, and if they are not there, then using the default values.

PR #654, opened 2022-01-07
fixed dependency install error on debian based

tuxutku:
apt was throwing error because valgrind:i386 has dependency clash with:  `libdrm-dev mesa-common-dev valgrind xserver-xorg-dev` (I'm using ubuntu). Fixed the issue by adding it to the 32bit depedencies


Tk-Glitch:
Thank you tuxutku! :frog:

PR #675, opened 2022-01-22
fix pipewire-jack not providing jack2 anymore on arch linux anymore

### LINKED ISSUE #671: Dependancy conflict arch linux. Wine-tkg wants jack2 but pipewire-jack is installed
Hi. Recently in pipewire based configuration jack2 was replaced with pipewire-jack. however pipewire-jack doesn't provide jack2 in dependancy resolutions as jack2-dbus can't   be used.
If wine doesn't expect the specific jack2 implementation using jack nd lib32-jack as dependancy instead of jack2 and lib32-jack2 can be the solution

ninelore:
Fixes #671 

I replaced jack2 with jack (same for lib32) in the PKGBUILD to counter a change on Arch linux where pipewire-jack doesnt provide jack2 anymore.

This change lets the user decide at buildtime to install either jack2 + lib32-jack2 or pipewire-jack + lib32-pipewire-jack if the user hasnt installed either. 

We probably should mention somewhere that pulseaudio users should pick former while pipewire users should pick latter


ninelore:
building rn. i will report back if this change causes issues

ninelore:
builds and works fine with `pipewire-jack`

Tk-Glitch:
I fear annoyances with chroot building, but it might be worth doing it now and find some workaround later on. Thanks ninelore! :frog:

PR #686, opened 2022-01-31
prepare: Allow _fsync_futex_waitv safelly ignore _use_fsync toggle

Iglu47:
Small addition to the applying of futex_waitv patches. Probably should be especially after https://github.com/Frogging-Family/wine-tkg-git/commit/764ff67d6fd7aad4edc3d03d7d46d69c6bbd07ff


Tk-Glitch:
Other patches depending on fsync would get disabled by allowing to simply enable futex_waitv without enabling fsync, so I have decided to enforce fsync to enabled when futex_waitv is enabled instead.

PR #720, opened 2022-03-26
Fix steamwebhelper patch

openglfreak:
This was broken before, sorry about that.


Tk-Glitch:
Thanks openglfreak! 🐸 ❤️

Issue #635, opened 2021-12-16
Ubuntu builds failed in actions


swanux:
Ubuntu builds have been failing for more than 20 days. A fix would be appreciated.

SveSop:
Probably nothing too big so that it cant be handled.. It is a multilib issue with gstreamer or someshit probably. (Mfplat req perhaps?)

I also see that libgcrypt20-dev and samba-dev are currently not added at all, so that should perhaps also be added, but it requires some libs to be manual replaced instead of using the meta-package name. Could be something similar with this error.

I can take a look at the deps when i get home, if its not fixed by then :)

SveSop:
I did not have the same issues with my ubuntu:focal docker image with the missing deps.. however i see that the buildlog indicates `42 packages can be upgraded. Run 'apt list --upgradable' to see them.` It COULD be some package dependency update happened that breaks this? My docker image is recent enough that it only had 1 upgradable package.
Dunno.. Was not able to recreate that with docker.
Maybe: `Run sudo dpkg --add-architecture i386 && sudo apt update && sudo apt upgrade` in the script to be sure?

If wine is to be built without missing deps for this:
```
configure: libusb-1.0 64-bit development files not found (or too old), USB devices won't be supported.
configure: libnetapi not found, Samba NetAPI won't be supported.
configure: libgcrypt 64-bit development files not found, GCRYPT won't be supported.
```
i propose the following fix for deps:
```
--- deps_old	2021-12-16 15:33:33.595141317 +0100
+++ deps_new	2021-12-16 15:39:09.429918070 +0100
@@ -6,7 +6,7 @@
   # 64-bit
   msg2 ""
   msg2 "Installing 64-bit dependencies for Debian-based distros..."
[...]

PR #747, opened 2022-04-19
Fix ubuntu CI building

CharlesThobe:
Fixes https://github.com/Frogging-Family/wine-tkg-git/issues/635 by removing all packages that have the string "deb.sury.org" in their version by using aptitude, this allows installation of 32-bit dependencies that were conflicting previously.

This PR probably should be reverted when ubuntu 22.04 github runner gets released.


CharlesThobe:
Update: I squashed the CI commits and better worded them.

Tk-Glitch:
I should have done that a while ago. Thanks for taking care of it! Ubuntu users will have an easier time! :frog: :heart:

CharlesThobe:
You're welcome!

PR #752, opened 2022-04-23
Update clock_monotonic patches

Iglu47:
My tests show that diff handles this and patches plain and staging correctly.
But this is mainly required to improve compatibility with fast_sync patches.


Tk-Glitch:
I think the initial issue was with futex2, but I can't remember exactly. It won't matter for recent trees where futex2 isn't a thing anymore. Also we should probably just drop that patch altogether on newer trees anyway :frog:

It's not really an issue as is, so let's merge! Thanks Iglu47! :heart:

PR #753, opened 2022-04-23
Update fastsync patches

Iglu47:
wine-side part of winesync/fastsync https://lkml.org/lkml/2021/1/17/312

source: https://repo.or.cz/wine/zf.git/shortlog/refs/heads/fastsync4
Esync/Fsync compatible version of fastsync, based on work @openglfreak

### DIFF
--- a/proton-tkg/proton-tkg-profiles/sample-external-config.cfg
+++ b/proton-tkg/proton-tkg-profiles/sample-external-config.cfg
@@ -140,6 +140,11 @@ _LOCAL_PRESET=""
 # ! Make sure to disable staging if your source isn't compatible with it or it will fail to apply !
 _custom_wine_source=""
 
+# fastsync - disable at runtime with WINE_DISABLE_FAST_SYNC=1 envvar - Set to true to enable winesync/fastsync support - https://repo.or.cz/wine/zf.git/shortlog/refs/heads/fastsync4
+# !! on plain: Disables esync / fsync support !!
+# !! on staging: Requires fsync support !!
+_use_fastsync="false"
+
 _use_esync="true"
 _use_fsync="true"
 # Allow making use of the futex2 kernel interface for fsync - Requires a patched kernel such as linux-tkg - https://gitlab.collabora.com/tonyk/wine/-/commits/experimental_5.13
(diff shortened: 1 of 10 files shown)


Tk-Glitch:
I'm still not convinced we want that builtin. Let's see how it goes I guess.

Thanks Iglu47! :frog: :heart:

Moonl1ghter:
Don't know if this is the place for it, but I was experimenting with fastsync a bit last week, and my experience is that the performance is identical to fsync. In theory it should cause less bugs than fsync does, what is the reason you don't want it @Tk-Glitch ?

Tk-Glitch:
It's slower than fsync overall still, and in key scenarios the difference can be pretty big. But more importantly kernel support is painful on distros not offering a dkms package for it (pretty much all non-pacman distros), and development from Zeb was on hold for a while due to blocking bugs.

It's not that I don't want it *at all* (more of the contrary), I'm simply not sure I want to provide actual explicit builtin support for it in the current state of things.

Now that development has been pushed forward again, I prefer to keep being careful to both not put any pressure on Zeb's shoulders and to make sure the issue tracker doesn't get overflowed by broken reports.

My plan was to offer it through the community patches repo after I disabled explicit builtin support with https://github.com/Frogging-Family/community-patches/commit/9fc318d0cee89ba4624fef686f77a32f833d4c7a but it needed special care to handle the patchset acceptably that way.

I'm also changing things quite a bit in an incoming update to the wine-tkg repo so adding more brittle parts to it makes me slightly nervous :D

Iglu47:
this change is required if you want to continue using `proton_LFH.patch`, but an error occurs when applying the patch
```diff
--- a/community-patches/wine-tkg-git/0002-proton_LFH.mypatch
+++ b/community-patches/wine-tkg-git/0002-proton_LFH.mypatch
@@ -2306,9 +2306,9 @@
      void              *param;         /* thread entry point parameter */
      void              *jmp_buf;       /* setjmp buffer for exception handling */
 +    void              *heap;          /* thread local heap data */
+     unsigned int       fast_alert_obj; /* linux object for the fast alert event */
  };
  
- C_ASSERT( sizeof(struct ntdll_thread_data) <= sizeof(((TEB *)0)->GdiTebBatch) );
 
 From 361dc9564a98d195ee8442d9f89acc5aa8e87c7b Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?R=C3=A9mi=20Bernon?= <rbernon@codeweavers.com>
 ```

PR #768, opened 2022-05-14
Update fastsync patches for e534d65 and move previous versions to legacy

Iglu47:
update due to https://github.com/wine-staging/wine-staging/commit/e534d6546a5be9f1cd53a0ea3ac79db7d977bed7


Tk-Glitch:
Thanks for keeping this up Iglu47! :frog: :heart:

PR #792, opened 2022-06-28
Added winex11: Moved child window buffer copy to vkQueuePresentKHR to…

### LINKED ISSUE #790: Vulkan child window buffer copy happens at the wrong point
Normally the copy is done in vkAcquireNextImageKHR ( [relevant line](https://github.com/Tk-Glitch/wine-tkg/blob/af025a6db4e8c7392a241c7747a11fc6ede52d77/dlls/winex11.drv/vulkan.c#L366) ) but it would make more sense to do it after a call to vkQueuePresentKHR. This causes a bug in SeriousEditor (and potentially other programs) where the ui is displayed with a 1 event delay   making it unusable (as well as showing uninitialized buffers whenever they are reinitialized). [This](https://github.com/pac85/wine-tkg/commit/0e4cb30a9e3b03c9ae66a20331b0f63b470642c1) is a quick fix I made.

Although less noticeable it could also cause a 1 frame input lag in games that make use of child window rendering, my patch should also fix it.

pac85:
Normally the copy is done in vkAcquireNextImageKHR ( [relevant line](https://github.com/ValveSoftware/wine/blob/a7618abea5ffeb3bfb1e69c0dbcdc1008bd88163/dlls/winex11.drv/vulkan.c#L445) ) but it would make more sense to do it during a call to vkQueuePresentKHR. This causes a bug in SeriousEditor (and potentially other programs) where the ui is displayed with a 1 event delay making it unusable (as well as showing uninitialized buffers whenever they are reinitialized).

Fixes #790

### DIFF
--- a/wine-tkg-git/wine-tkg-patches/misc/childwindow/childwindow-proton.patch
+++ b/wine-tkg-git/wine-tkg-patches/misc/childwindow/childwindow-proton.patch
@@ -959,3 +959,91 @@ index 00000000000..3ab31201d3d
 +#endif /* defined(SONAME_LIBXFIXES) */
 +
 +#endif /* __WINE_XFIXES_H */
+From 5ebbeb4784aa9e67f92a295eda46ee1c74b9a1f2 Mon Sep 17 00:00:00 2001
+From: antonino <antomani103@gmail.com>
+Date: Tue, 28 Jun 2022 22:04:04 +0200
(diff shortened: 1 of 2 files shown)


Tk-Glitch:
Thanks again @pac85 ! Let's merge :)

pac85:
Thank you again for giving me the opportunity to contribute!

PR #821, opened 2022-08-07
Hotfixes: mfplat: Restore mfplat streaming support on 7.2+

Iglu47:
"Upstreaming of the necessary patches is almost complete, so we will take them directly from vanilla Wine-7.14+
The implementation of these patches is slightly different from the previous one (last used on [Wine-Staging-7.2+](https://github.com/wine-staging/wine-staging/tree/bcfed21ea1925e06c1f0db0e86cb2380300b8aa9/patches/mfplat-streaming-support)). It is currently used in the `experimental_7.0` branch of Proton.

Affects version https://github.com/wine-staging/wine-staging/commit/05ea40b9366e4ee28a2baa461853b778230cbad1 and above. Should allow some people to stop holding on to wine-tkg-staging-7.2.r9.gbcfed21e

Also during testing, I found that somewhere after Wine-6.21+ movies in some games now depend on shared gpu resources. Probably now something triggers some game engines to use them. The same for the previous implementation, so it's not due to a difference in implementations. However, `experimental_7.0` contains these required [shared gpu resource patches](https://github.com/ValveSoftware/wine/commits/3d367c1bb0d48dd803fc0a8029c26c1b5cc3902f) so these videos will continue to work for it (with the required DXVK versions)."

@Tk-Glitch also it introduce initial utilize `git cherry-pick` (which probably needs to some rework). But there may be occasions for small fixes where you find it useful.

Should supersede https://github.com/Frogging-Family/wine-tkg-git/commit/b3598cf9482360547294c71daaa21e95c8fc6d2d and https://github.com/Frogging-Family/wine-tkg-git/commit/cb78cd59568d4191ba23474cec722e8544aae9fb so I removed these changes.

And of course will need to track when MRs are accepted into Upstream and add the maximum versions of Wine to which this fixes applies.


Iglu47:
fix some typo in descriptions

Tk-Glitch:
I'm not sure we'll use the cherry picker very often, but that could be useful in the future indeed. The need for reworking it will likely only come if we start using it more extensively.
Really nice work with that MR Iglu47! Thank you very much! :frog: :heart:

PR #822, opened 2022-08-07
Update the a2395ecf hotfix with Rémi's pending MR patch

Iglu47:
small syncing with MR


Tk-Glitch:
Thanks Iglu47! I got lost with that MR :frog:

PR #823, opened 2022-08-07
Hotfixes: Add workaround for Guild Wars 1 and Guild Wars 2 login issue on 7.14

Iglu47:
fixes connection issue for Guild Wars 1 and Guild Wars 2 with patch from Paul Gofman
more info: https://bugs.winehq.org/show_bug.cgi?id=52429


Tk-Glitch:
I see it was largely simplified over the first iterations. Nice oneliner :frog:
Thanks again Iglu47! :frog: :heart:

Iglu47:
yes, as I understand it, there is no such problem on experimental_7.0. and all because of some difference in versions and fresh patches - https://github.com/wine-mirror/wine/commit/9783e2e16c35586b5fa65ae27ac1f96c6260c534

PR #828, opened 2022-08-09
Hotfixes: mfplat: Disable on 7723063e+

Iglu47:
https://gitlab.winehq.org/wine/wine/-/merge_requests/582 merged

this should already work in Upstream, but let's keep these patches together


Iglu47:
since I removed https://github.com/Frogging-Family/wine-tkg-git/commit/b3598cf9482360547294c71daaa21e95c8fc6d2d let's give `restore_staging_mfplat` new boundary

Tk-Glitch:
Thanks Iglu47! I'll change the boundary so it includes the first breaking commit instead of the tail. It's unlikely someone will target the specific commit that will fail, but still... :frog:

Iglu47:
ok, I'll take it as rule

PR #830, opened 2022-08-11
fastsync: Improve compatibility with sharedgpures

Iglu47:


Tk-Glitch:
Quite the context reduction. Hopefully that won't bite us when we least expect it 😄 
Thank you Iglu47! :frog: :heart:

PR #835, opened 2022-08-18
sharedgpures: Improve compatibility with fullscreen hack

Iglu47:
split sharedgpures.patch and add fshack version


Tk-Glitch:
Looks good! Thanks Iglu47! That will make quite a few users happy :frog: :heart:

PR #847, opened 2022-08-26
sharedgpures: Discard avoidable reverts

Iglu47:
Sorry, I couldn't take the time to resolve it properly right away, but @Guy1524 explained that it's enough to just resolve the conflicts. Less code, should be more maintainable.

Also sharedgpures-fixup is not intended to contain absolutely all Media Foundation source related fixes, but only some proton-specific CW-Bug-Id: **#19126** patches, which probably cannot be accepted in the current state to Upstream, but since https://github.com/wine-mirror/wine/commit/76db4b2459920eb2e2004255e3388b069dd1d6f9 already exists, let em be in sharedgpures-fixup-76db4b2


Tk-Glitch:
Thanks once more Iglu47! 🐸 ❤️

PR #849, opened 2022-08-29
sharedgpures: Fix shared gpu resources patchset

Iglu47:
This part of the old code was mistakenly taken from deprecated branch https://github.com/Guy1524/wine/tree/shared-resources and now interferes.

Fixes the launch of probably most Vulkan games. (Hades [vulkan] still broken like on current `experimental_7.0`.)


Tk-Glitch:
Thanks Iglu47! :frog: :heart:

Issue #860, opened 2022-09-09
Shared GPU resources patchset doesn't apply


llde:
`sharedgpures-textures.patch -- ( Applied shared texture resources support patchset )
patching file dlls/vulkan-1/tests/vulkan.c
patching file dlls/winevulkan/make_vulkan
Hunk #1 succeeded at 111 with fuzz 2.
Hunk #2 succeeded at 126 (offset 1 line).
Hunk #3 succeeded at 191 (offset 1 line).
Hunk #4 FAILED at 199.
Hunk #5 succeeded at 237 (offset 1 line).
Hunk #6 succeeded at 267 (offset 1 line).
Hunk #7 FAILED at 1103.
Hunk #8 succeeded at 1269 (offset 7 lines).
Hunk #9 succeeded at 2203 (offset 7 lines).
Hunk #10 succeeded at 2345 (offset 7 lines).
2 out of 10 hunks FAILED -- saving rejects to file dlls/winevulkan/make_vulkan.rej
patching file dlls/winevulkan/vulkan.c
Hunk #1 FAILED at 24.
Hunk #3 succeeded at 383 (offset 8 lines).
Hunk #4 succeeded at 398 (offset 3 lines).
Hunk #5 succeeded at 1306 (offset 3 lines).
Hunk #6 succeeded at 1365 (offset 3 lines).
Hunk #7 succeeded at 1378 (offset 3 lines).
Hunk #8 FAILED at 1963.
2 out of 8 hunks FAILED -- saving rejects to file dlls/winevulkan/vulkan.c.rej
patching file dlls/winevulkan/vulkan_private.h
Hunk #1 succeeded at 226 (offset 22 lines).
patching file dlls/winevulkan/make_vulkan
Hunk #1 FAILED at 204.
Hunk #2 succeeded at 280 (offset -2 lines).
1 out of 2 hunks FAILED -- saving rejects to file dlls/winevulkan/make_vulkan.rej
patching file dlls/winevulkan/vulkan.c
Hunk #1 succeeded at 1383 (offset -11 lines).
Hunk #2 succeeded at 2000 with fuzz 2 (offset -289 lines).
patching file configure.ac
Hunk #1 succeeded at 3075 (offset 4 lines).
patching file dlls/sharedgpures.sys/Makefile.in
patching file dlls/sharedgpures.sys/shared_resource.c
patching file dlls/sharedgpures.sys/sharedgpures.sys.spec
[...]

PR #861, opened 2022-09-13
Rebase sharedgpuresources patchset  for Wine 8.0

llde:
To simplify the rebase process for myself, I did some rework on the patch organization:
Created a sharedgpures-driver.patch that contain all changes related to the sharedgpures.sys driver and releant INI/Registry changes. Also the ntoskrln changes were added here. 
Added all multimedia related changes (winegstreamer/mf) to sharedgpures-fixup.patch

As rebasing the sharedgpures-texture and sharedgpures-fences patchset directly where problematic I remade the changes in the source and recreated the patches. This cause that all the code changes are squashed into a single commit.
Old code assumed that Vk*** types were strctures, they were actually changed to handles, so where necessary I added the relevant handle to type functions, and changed the code to use structure directly in the internal functions.

I also readded the control of supported extensions in device creation, it properly error when it's requested an extension that is  unsupported by the device, with the exclusion of the extensions with a special handling added by the patchset.
These 2 patches now contain only winevulkan changes.

FSHACK version is still not reworked. I don't use fshack (break some game with small resolutions as Imperium GBR), the included file is however stripped of the changes that went to driver or fixup subpatches.

I did test it and it works as it doesn't seems to break applications. Tested d3d9 and d3d11.  Didn't test the added functionality as I don't have application that use them .

 Thanks and Asseghimma. 

Close bug #860


kolAflash:
Works fine for me!

Also for running [Warcraft III Reforged 1.33.19308](https://github.com/doitsujin/dxvk/issues/1866#issuecomment-1220034968).

CartoonFan:
@llde Sorry for asking, but will this work with #862?

Iglu47:
> Sorry for asking, but will this work with #862?

ofc no. I was forced to made #862 because this #861 breaks shared fences and I don't have a better alternative at this moment also. Thanks to @llde for looking at this anyway.

llde:
@Iglu47  If I'm able to start Control  in DX12 I will check and try to fix them.

llde:
@Iglu47  last commit solve the issues with Control and Virtual Queues for me.

Tk-Glitch:
Thanks @llde, I'm in for merging.

@Iglu47 With these last fixes from llde, is there something still needing a fix here? FS hack specific reverts from 862 would be the only thing left, right?

llde:
@Tk-Glitch there is a reason why FSHack cannot be rebased to current master, and instead rely in an ever increasing revert list?

Iglu47:
fshack version isn't broken since https://github.com/Frogging-Family/wine-tkg-git/commit/1bbdaa4c563e476f9022f05a55d7b237378fdbb9#diff-7c3f6ce9ee8af2ace2f529d77b9320dd8fc6fb68bceaac2790ab778decc32bf5R169 already contains requared reverts from https://github.com/Frogging-Family/wine-tkg-git/pull/862

In this form, it's hard for me to say how many changes compared to the original patches and whether it's safe to say that problems with the new sharedgpures patchet can be addressed directly to Derek.
Apparently these patches can be merged in such form for now (I will revise it later).

Iglu47:
> last commit solve the issues with Control and Virtual Queues for me.

Yes, for me too. Great job!

> With these last fixes from llde, is there something still needing a fix here?

I don't really have many applications where I could check the operation of each part of the patchset. This version is currently being successfully tested on applications available to me (I have only the Control for shared fences and the Unity game for textures.)

> FS hack specific reverts from 862 would be the only thing left, right?

no, 862 is an alternative that would only put the patchset on the path of reverts (similar to fshack)

I don't agree with how the patches are now split, but since it works I certainly don't mind accepting it now.
however, I will reworking splitting back to the old one soon (if I have to continue to maintain this)

llde:
> 

> I don't agree with how the patches are now split, but since it works I certainly don't mind accepting it now.
> however, I will rework splitting back to the old one soon (if I have to continue to maintain this)

Why?  The driver part is the one that would overall require way less rebases for example, it make more sense to me to keep it separate. The same with the multimedia changes, the rebases for this are indipendant on the proper vulkan side. 
As it were the texture and the fence patches contain winegstreamer and mf changes,, so a breaking change in these component would require a rebase of 3 patches. Now it require the rebase of just 1 patch.

Maybe you are thinking of something I'm not seeing, but this seems like a better condition to me.  I would like to know your mind on this.

Iglu47:
Patches were split not to save space, but because of functionality. Unlike most other patchsets, sharedgpures conditionally consists of two parts: a completely independent textures part and a more complex fences part - personally I would like to keep the ability to quickly disable and test the patchset without the more problematic heavy part (yes, we don't have flags for this, but clear commits group separation allows it).

llde:
We still can disable the fence patchset, then multimedia changes present in there were quite trivial and even if they are appllied I don't see how your use case have issues.  
Actually the only multimedia related change is :
winegstreamer: Add MF_MT_VIDEO_NOMINAL_RANGE attribute to base video output type.
that seems something that can easily live stand alone.

Some commits related to the sharedgpures.sys do fix some issues introduced witht the texture patchset, even if these issues aren't triggered. I don't see why they should be on a separate patchset.
The bulk of the fence patchset is winevulkan changes after all. 
The mapping changes are imho the only changes that could need proper attention. 
We could however move these changes into a driver-fence patch instead of using again 3 bulk mad patches.

The only things is that you could disable mulltimedia changes without the texture and fences patchset. However this is already na issue, if the stuff in the fixup patchset are required, as it can be disabled without the others. I just added in fixup the patches that were in texture patchset and the only one in the fence one. (I mean most of these are trivial and simple changes)
I also believe that these patches could stand alone. Paradoxically the multimedia changes that seems most realted to the shared gpu resource functionality are in fixup (the ones from Giovanni Mascellani).

Tk-Glitch:
> @Tk-Glitch there is a reason why FSHack cannot be rebased to current master, and instead rely in an ever increasing revert list?

It needs to be rewritten at this point as there were too many changes upstream (the bulk of it is behind but there are still some fixes dropping every couple days). The rewrite will be published with the next Proton major release and we'll offer it in -tkg around the same time, depending on how things will go. We might offer the updated patchset before the Proton update drops, but no promise.

Tk-Glitch:
> Patches were split not to save space, but because of functionality. Unlike most other patchsets, sharedgpures conditionally consists of two parts: a completely independent textures part and a more complex fences part - personally I would like to keep the ability to quickly disable and test the patchset without the more problematic heavy part (yes, we don't have flags for this, but clear commits group separation allows it).

Such "flags" could be added to make things easier as needed. Lower maintenance burden deserves such an addition imo.

llde:
Started working again on updating the patchset for current master.
Recently staging added a discrepancy in the multimedia subpatches, requing a different fixup patchset. (For now only staging is implemented,  non staging will give an error)

Still doesn't work. While I rebased the texture patchset, it doesn't compile properly in 32 bit inside one of the functions generated by make_vulkan script. I'm going to take a better look tomorrow.

llde:
Ok patchset rebased for staging commit 49ed7ae4a001a5d84141f65064081cce3c136137  and related mainline commit with added sharedgpuresources-fixup patchset for mainline.

Considering the sheer amount of changes, in particular hand-rolling the chain conversion functions (make_vulkan doesn't generate them anymore, it instead hand roll them of the conversion inside the thunk, so it happens before the PE function call), stubbing them from  make_vulkan script. (It was also generating wrong code).
Handrolling the functions to clone and fixup the struct chain allow us to remove some affecting code from make_vulkan regarding the dynamic arrays deep copy for VkSubmitInfo and VkSubmitInfo2.  

the Unexposed extensions mechanism changed. Before it was generating the function native entry point in the device function list
 , but after a specific commit upstream doesn't anymore. So the extensions in the list were commented out.

I took some liberty on rewriting some code, (removed the double loop in wine_vk_device_convert_create_info for example) also avoiding some changes introduced in the texture patchset and changed back with modifications in  the fence patchset. 

Good part of the modification is also related to manually convert the wine handles to native handles beofre calling the unix function. Before the entry points from the thunk entry was the wine_* impl and the conversion was probably happening somewhere else. Now the script generate a thunk entry for every function, even the ones with a wine_ impl.

The patchset was tested with Control in DX12 mode and in Dx11 mode with DXVK (a recent version some commits earlier then official 2.0). It works fully, however I crash in vkWaitSemaphoresKHR (assertion  failed in wine)  when trying the DX12 mode with vulkan validation layers active, but the validation layer say nothing. (Dx12 use vkQueueSubmit while DX11 with DXVK use vkQueueSubmit2 so both were tested)

While I was making this, wine introduced yet other changes for winevulkan, so the patch isn't applicable at last upstream, or staing rebase. These changes should be however minor, and the fact this now handroll it's own conversion functions, it doesn't rely on correctness in generating the code anymore.

Saancreed:
Hi @llde, are you still working on this? Wine is currently in code freeze so for the next few weeks there should be little changes to winevulkan code, making this a good time to rebase the patchset once again.

llde:
@Saancreed when I have time. Understanding how different commits affect the pathset isn't trivial

llde:
Readded patch for wine 7.20 and added a rebase for Wine 8 (works for mainline and staging)

Tk-Glitch:
Awesome work by @llde here. Thank you !
It's time to merge :frog:

To people testing this out: feel free to give some feedback 😉

PR #862, opened 2022-09-13
sharedgpures: Add revert list

Iglu47:
sorry, the fences part seems to require non-trivial changes and it's quite difficult (at least for me), so some reverts of winevulkan are probably needed until the proper solution would be found

closes https://github.com/Frogging-Family/wine-tkg-git/issues/860

also minor changes for fences part: fix compiler warning for trace and remove unnecessary code from fshack version for non-fshack version.


Iglu47:
superseded by https://github.com/Frogging-Family/wine-tkg-git/pull/861

Issue #846, opened 2022-08-25
[Proton-tkg Script] Replacing Old Builds Mis-prioritizes New Replacements


ThisNekoGuy:
Replacing old builds with new ones prioritizes the highest number in the list rather than the build just made:
```
####################################################################################################

 Proton-tkg build installed to /home/neko-san/.steam/root/compatibilitytools.d/proton_tkg_7.15.r5.gb31a964e.Staging-futex_waitv

####################################################################################################

Do you want to run the uninstaller to remove previous/superfluous builds? N/y: y
Your config.vdf file was backed up from /home/neko-san/.local/share/Steam/config/config.vdf (.bak)

What Proton-tkg build do you want to uninstall?
  1 - 7.2.r9.gbcfed21e.ASYNC-Staging-futex_waitv
  2 - 7.14.r0.g1d21e2b6.ASYNC-Staging-futex_waitv
  3 - 7.14.r0.g1d21e2b6.Staging-futex_waitv
  4 - 7.15.r5.gb31a964e.Staging-futex_waitv
  5 - experimental.bleeding.edge.7.0.23049.20220821.ASYNC-ValveBleeding-futex_waitv
choice [1-5]: 3
###########################################################################################################################

Proton-tkg 7.14.r0.g1d21e2b6.Staging-futex_waitv was uninstalled and games previously depending on it will now use Proton-tkg experimental.bleeding.edge.7.0.23049.20220821.ASYNC-ValveBleeding-futex_waitv instead.

###########################################################################################################################

```

Tk-Glitch:
That's actually by design, but yeah that may be better

ThisNekoGuy:
Think this edit would work?
[proton-tkg.sh.txt](https://github.com/Frogging-Family/wine-tkg-git/files/9579497/proton-tkg.sh.txt)

ThisNekoGuy:
Ah, seems that it does; I'll do a PR

ThisNekoGuy:
Is... the PR going to get reviewed?

PR #864, opened 2022-09-15
Updated proton-tkg.sh Uninstaller for Issue #846

ThisNekoGuy:
Updated proton-tkg.sh uninstaller to replace selected builds with the one just made rather than whatever has the longest string-name.
Namely, the issue happens when Valve builds are installed.


ThisNekoGuy:
Fixes:

- https://github.com/Frogging-Family/wine-tkg-git/issues/846

Tk-Glitch:
Thank you. This breaks standalone usage of the uninstaller (which is the only way I'm using it myself), so I won't merge as is. I'll rework it asap but it's low on my list let's be fair. I should do it pretty soon still since it'll be pretty small.

PR #867, opened 2022-09-26
Don't specify the exact version for some deps

gabriele2000:
Fixes "cannot find: <name package>" of some deps because they're newer than the expected.


Tk-Glitch:
Thanks gabriele2000! :frog: :heart:

PR #884, opened 2022-11-08
Wine-tkg-git: Build: Fix PE files in the same way as proton-tkg builds.

logos128:
Makes PE files compatible with Forza Horizon games, etc.

I've been using this for several months now without any issues.

### DIFF
--- a/wine-tkg-git/wine-tkg-scripts/build.sh
+++ b/wine-tkg-git/wine-tkg-scripts/build.sh
@@ -278,10 +278,26 @@ _package_nomakepkg() {
 	cp -v "$_where"/wine-tkg-scripts/wine-tkg-interactive "$_prefix"/bin/wine-tkg-interactive
 
 	# strip
-	if [ "$_pkg_strip" = "true" ] && [ "$_EXTERNAL_INSTALL" != "proton" ]; then
-	  for _f in $( find "$_prefix" -type f '(' -iname '*.dll' -or -iname '*.so' -or -iname '*.sys' -or -iname '*.drv' -or -iname '*.exe' ')' ); do
-	    strip --strip-unneeded "$_f" && msg2 "$_f stripped"
-	  done
+	if [ "$_EXTERNAL_INSTALL" != "proton" ]; then
+	  if [ "$_protonify" = "true" ] && ( cd "${srcdir}"/"${_winesrcdir}" && ! git merge-base --is-ancestor 2e5e5ade82b5e3b1d70ebe6b1a824bdfdedfd04e HEAD ); then
+	    if [ "$_pkg_strip" = "true" ]; then
+	      msg2 "Fixing x86_64 PE files..."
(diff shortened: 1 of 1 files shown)


wine-tkg-git/wine-tkg-scripts/build.sh
```diff
@@ -278,10 +278,20 @@ _package_nomakepkg() {
...
 
 	# strip
-	if [ "$_pkg_strip" = "true" ] && [ "$_EXTERNAL_INSTALL" != "proton" ]; then
-	  for _f in $( find "$_prefix" -type f '(' -iname '*.dll' -or -iname '*.so' -or -iname '*.sys' -or -iname '*.drv' -or -iname '*.exe' ')' ); do
-	    strip --strip-unneeded "$_f" && msg2 "$_f stripped"
-	  done
+	if [ "$_EXTERNAL_INSTALL" != "proton" ]; then
+	  if [ "$_pkg_strip" = "true" ]; then
+		msg2 "Fixing x86_64 PE files..."
+		find "$_prefix"/"$_lib64name"/ -type f -not '(' -iname '*.pc' -or -iname '*.cmake' -or -iname '*.a' -or -iname '*.la' -or -iname '*.def' -or -iname '*.py' -or -iname '*.pyc' -or -iname '*.pl' ')' -printf '--strip-unneeded\0%p\0%p\0' | xargs -0 -r -P1 -n3 objcopy --file-alignment=4096 --set-section-flags .text=contents,alloc,load,readonly,code
```
ViNi-Arco, line 284: This "appears" to break strip, wouldn't it be "strip --strip-unneeded" instead of just "--strip-unneeded"?
same for lines 286, 287, 427, 429, 430.

ViNi-Arco (changes requested):

wine-tkg-git/wine-tkg-scripts/build.sh
logos128: It strips the binary via objcopy along with the file alignment and the .text section flags. The -printf part ('--strip-unneeded\0%p\0%p\0') is being added by xargs at the end of the objcopy command.
Proton does the same.

ViNi-Arco: Yes, I can see in the [documentation](https://sourceware.org/binutils/docs-2.38/binutils/objcopy.html) that it is possible, interesting that you can use **objcopy** for that, I'm used to **Binutils** for strip

Tk-Glitch:
Yeah I don't see an issue with that except for the fact we might want to make it optional (enabled by default is fine though). It could change behavior on plain builds which is undesirable to report upstream.

ViNi-Arco:
>we might want to make it optional

Good idea, separating this new behavior from the old one as an option..

logos128:
> Yeah I don't see an issue with that except for the fact we might want to make it optional
I agree. IMO we can use the _protonify flag as it seems appropriate for this.
BTW, upstream just fixed this issue with https://github.com/wine-mirror/wine/commit/2e5e5ade82b5e3b1d70ebe6b1a824bdfdedfd04e . So we finally have correct flags for the .text section of PE binaries :)
We can condition the use of objcopy to this commit as well. (If I'm not mistaken there is a patch by rbernon in proton-tkg.patch, that sets the file alignment to 4096 at compile time.)

logos128:
I modified the commit with the above proposals.

PR #885, opened 2022-11-08
Protonify: Fix some issues and add a few additional patches mostly needed for Forza Horizon games.

logos128:
Allows running Forza Horizon 4/5 with wine-tkg.

A few patches from the FS hack (although not directly related to that functionality) rebased to the latest Wine tree and added as an additional patch. The patch only applies on _protonify and inactive _proton_fs_hack. That way the games can be used without the FS hack and also avoid all the reverts :).


Tk-Glitch:
There are only windows store and steam versions of those games, right? I'm not too comfy with this but we already allowing to do this with Valve trees so heh. Thanks logos128! :frog:

PR #892, opened 2022-11-14
sharedgpures: Update sharedgpures patches for wine-7.18

Iglu47:
- small update for 7.18 (https://github.com/wine-mirror/wine/compare/7384591c71788bc99a3e4be48bfc09d1139e5243...f664af0bca4af1909bd30bc6da1e116d1cd90f9a)
- fix 7.14
- concatenate textures and fences parts (stop segmentation, because utilize for fshack is not possible due to reverts).


llde:
Does more recent wine versions introduce regressions that make targetting 7.18 somewhat valuable or the plan is to have  a version of the patchset for at least every  release?

Iglu47:
Usually the goal is to have a versions of the patchset for each group of commits accepted in Wine on at least one day. Unfortunately, I cannot promise that I will do all this. I may not have that much free time or (more critically) skill level.

llde:
Recently from the wine side they changed a lot of things. All the functions start from the thunks and only after the wine impl is called. They removed struct chain conversion functions, instead generating them in the win-host conversion functions. Broke them, flat out generating incorrect code if there are not pChain elements that must be removed, and the orginal chain cannot be passed to the PE implementation anymore. This break functions that require to see the old chain to be implemented properly.
In #861 I rebased the patchset for 7.20 (commit of a 4-5 days before 7.21 actually), keeeping the structure I'm more familiar to work. If you want to take a look.

I see that you abandoned the multi file patch idea.  Wouldn't be better to at least keep the multimedia related changes separated? 
(I mostly think that these changes were insturmental on allowing some games to work but not related to the good working of the shared resources themselves ) They should be mostly the same wither with fshack or without.

Iglu47:
- separate media-related changes from the main file.
- explicitly specify the range of supported versions, as we are aware of the issue.

Iglu47:
just checked on 7.18: RT in Crysis: Remastered works on my machine in the same way as on Proton experimental_7.0

Tk-Glitch:
Looks good to me. Thank you @Iglu47 and @llde ! Let's frog. :frog: :heart:

PR #894, opened 2022-11-16
Remove Ubuntu Docker notice from wine-tkg-git/

orowith2os:
Link is outdated, or is there some other link that should be used?


Tk-Glitch:
It's gone indeed. Thank you ! :frog: :heart:

PR #895, opened 2022-11-17
[wine-tkg] Call tool/make_makefile after userpatch application, plus a autoconf fix

llde:
Hi. As  I was writing on discord yesterday, currently wine-tkg doesn't call tools/make_makefiles however it seems to include at least one patch that cause doesn't properly include the configure.ac changes. 
This can be seen after the preparation step by using the following commands in the wine_mirror-git folder
git add * (to add all files to git )
git commit -m "base"  (create a base commit)
./tools/make_makefiles
git diff

It's possible to see various changes around nv* dlls (this is with staging, protonify, sharedgpures), a sign it wasn't updated properly before

The change in this PR also allow to use properly patchset that doesn't properly provide configure.ac modifications for added or modified dlls. And also allow  is to remove most of the configure/configure.ac parts of the patchsets.

Linked there is the other issue.

To fix opencl.h the prepare.sh sciprt is doing a sed pass to replace OpenCL/opencl.h to CL/opencl.h in configure and configure.ac (to avoid autoconf issues), however the second changes have issues.
AC_CHECK_HEADERS use the name of the header to generate the cache var and the define. With OpenCL/opencl.h it was generating:
HAVE_OPENCL_OPENCL_H as a define guard
ac_cv_header_OpenCL_opencl_h as a cache var of the result

The first is used inside wine source and the second is used later in the configure.ac script itself
Now when I use autoreconf with the new header value it change how these are generated, generating:
HAVE_CL_OPENCL_H as a define guard
ac_cv_header_CL_opencl_h as a cache var of the result

however both are senseless in the optic of the configure script and wine code.

The patch is more complex and exact:
It remove the header check from the AC_CHECK_HEADERS and use AC_CHECK_HEADER to make  a check for this specific header. When the header is found I explictily  define the guard HAVE_OPENCL_OPENCL_H
the other script part where it was using ac_cv_header_OpenCL_opencl_h  are changed to use ac_cv_header_CL_opencl_h
(as what variable it generate for the result cannot be changed)

Now autoreconf doesn't break opencl anymore

### DIFF
--- a/wine-tkg-git/wine-tkg-patches/hotfixes/autoconf-opencl-hotfix/hotfixes
+++ b/wine-tkg-git/wine-tkg-patches/hotfixes/autoconf-opencl-hotfix/hotfixes
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+# Patch opencl.h header to it's correct location without breaking successive autoconf call, as it was done by the sed workaround
+if ( cd "${srcdir}"/"${_winesrcdir}" ); then
+  warning "Hotfix: Fix for CL/opencl.h header recognition"
+  _hotfixes+=("$_where"/wine-tkg-patches/hotfixes/autoconf-opencl-hotfix/opencl-fixup)
+fi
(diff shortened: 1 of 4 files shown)


Tk-Glitch:
Looks fine to me. Thank you llde ! :frog: :heart:

PR #899, opened 2022-11-20
Add explanation why nightly builds can't be downloaded for users that are not logged in

Elendil211:
An option to download nightly builds without logging in would be better, but in this way there is at least an explanation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Wine to rule them all !
 
+You must be logged in to GitHub in order to download Wine or Proton nightly builds.
+
 ## Wine nightly builds
 
 - wine-staging patchset applied


teppyboy:
You can download nightly build without logging in by using https://nightly.link. For example I want to download https://github.com/Frogging-Family/wine-tkg-git/actions/runs/3512850046, I can copy-paste that link into the website and it'll give the [downloadable artifact url without logging in](https://nightly.link/Frogging-Family/wine-tkg-git/actions/runs/3512850046/wine-tkg-build.zip)

Tk-Glitch:
I'm fine with this. Not sure if we want to link to the website teppyboy mentioned but we might. Thanks Elendil211 ! :frog: :heart:

PR #921, opened 2022-12-09
Implement single-make mode

openglfreak:
This allows building 64-bit and 32-bit Wine in parallel, using GNU Make's jobserver to limit the number of processes.

Reason: Some parts of Wine (e.g. actxprxy_shobjidl_p.c) take quite long to compile with high optimization flags, it would be nice if those parts wouldn't effectively halt the compilation process.

The name of the variable may be a little bit wrong, since we are actually starting multiple `make` processes, but there's only one jobserver. Eh, whatever.


Tk-Glitch:
Funny optimization experiment! Thanks openglfreak :frog: :heart:
It didn't make much of a difference on my main machine but I'll definitely give it a while on something slower.

PR #923, opened 2022-12-10
Fix the gitignore not ignoring wine-tkg-git/src

orowith2os:
This might affect other directories too, but I'm only interested in fixing this one.

If I find any other directories that need to be ignored, I'll make another MR including them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@
 *.old
 *.db
 *.files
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
Right, it defaults to a symlink these days. Thanks ! :frog: :heart:

PR #935, opened 2023-01-11
fastsync: Add some "QoL things" for testers

Iglu47:
- "immediately prevent building without winesync header"
if `_use_fastsync="true"` was set, we probably don't want to run into fastsync not working on _that_ build of Wine. No need to waste time compiling such builds.
- "not make conflicts fail-safe, stop the build immediately"
The same as the previous one. With `_use_fastsync="true"` we want the Wine build with Fastsync. So let's save the User time and let them start resolving conflicts _before_ compiling.
- "output terminal message if `/dev/winesync` not exist"
Sometimes people forget that they haven't loaded the winesync module. _Numerous_ complaints from fastsync should clear things up _a bit_.
- "make priority over fsync/esync (it's **NOT** failsafe, like fsync!)."
any `WINEESYNC` and `WINEFSYNC` values will be completely ignored untill we set env var `WINE_DISABLE_FAST_SYNC=1`. If `_use_fastsync="true"` was set, we're probably interested in getting this to work out of the box. Will help people to know for sure that they didn't mess up anything in the environment variables.

Not perfect, but should help with some common problems.

### DIFF
--- a/proton-tkg/proton-tkg-profiles/sample-external-config.cfg
+++ b/proton-tkg/proton-tkg-profiles/sample-external-config.cfg
@@ -156,9 +156,11 @@ _use_GE_patches="true"
 # ! Make sure to disable staging if your source isn't compatible with it or it will fail to apply !
 _custom_wine_source=""
 
-# fastsync - disable at runtime with WINE_DISABLE_FAST_SYNC=1 envvar - Set to true to enable winesync/fastsync support - https://repo.or.cz/wine/zf.git/shortlog/refs/heads/fastsync4
-# !! on plain: Disables esync / fsync support !!
-# !! on staging: Requires fsync support !!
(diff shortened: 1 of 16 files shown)


Iglu47:
TODO: "Tomorrow" I need to check legacy versions. And I'm not sure about where different distributions store the Linux header.

Iglu47:
@Tk-Glitch I think now you can take a look

Iglu47:
Add post-build message. Update description in .cfgs

Iglu47:
With changing the priority of `WINE_DISABLE_FAST_SYNC` over esync and fsync
https://github.com/Frogging-Family/wine-tkg-git/commit/400d5989e1da105d68df0787576a962204931b5c is probably no longer necessary. So let's revert it.

Tk-Glitch:
For proton builds, we might want to add the `WINE_DISABLE_FAST_SYNC` option to `wine-tkg-git/proton-tkg/proton_template/conf/user_settings.py` since we can't just enable ESYNC/FSYNC back.
Otherwise it looks fine to me.

Iglu47:
`PROTON_NO_NTSYNC` was added.

Iglu47:
`PROTON_NO_NTSYNC` was added (keep style like other and avoid spam to non-ntsync build by default)

PR #941, opened 2023-01-18
fsync: Rename toggle options

Iglu47:
enough time has passed, so there should be no more problems associated with the transition to using futex_waitv.
this PR should clean up some the fsync things:
- for [the main fsync patch script](https://github.com/Frogging-Family/wine-tkg-git/blob/c15b9c448bf101a3bf32bf15ee852199e5d430d1/wine-tkg-git/wine-tkg-patches/proton/fsync/fsync): more explicitly split patches into two blocks - before and after 5.20 (for versions before 5.20, bring checks to the current common style.)
- `_fsync_futex_waitv` already enabled by default since https://github.com/Frogging-Family/wine-tkg-git/commit/764ff67d6fd7aad4edc3d03d7d46d69c6bbd07ff . So let's just stick this to fsync-unix patches as the official fsync version (for wine-5.20+) and remove unnecessary checks. 
 - allow build (for wine-5.20+) with Fsync legacy (`FUTEX_WAIT_MULTIPLE`) and deprecated (`futex2`) versions
 
##### * _For older Wine versions (prior to 5.20), there is only one "Fsync Legacy" version, so there is nothing left but to continue to apply this version_

### DIFF
--- a/proton-tkg/proton-tkg-profiles/legacy/legacy-options.cfg
+++ b/proton-tkg/proton-tkg-profiles/legacy/legacy-options.cfg
@@ -76,8 +76,12 @@ _proton_mf_hacks="false"
 
 _plasma_systray_fix="false"
 
+# Disable futex_waitv patches and keep fsync legacy (FUTEX_WAIT_MULTIPLE opcode 31) on Wine 5.20 and above - https://steamcommunity.com/app/221410/discussions/0/3158631000006906163/
+_fsync_legacy="false"
+
 # Allow making use of the futex2 kernel interface for fsync - Requires a patched kernel such as linux-tkg - https://gitlab.collabora.com/tonyk/wine/-/commits/experimental_5.13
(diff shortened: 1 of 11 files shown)


Iglu47:
should be fine (reports about any found combinations of cfg options in which PR breaks something are welcome).
also I didn't find a reason why the option to disable futex_waitv patches would be useful.

Tk-Glitch:
Looks fine to me. Do you want to do further changes/additions to this or are you ready to merge?

Iglu47:
sorry for the delay on my part.
today I am doing the last checks to make sure that everything is fine and I did not miss anything

Iglu47:
now `_fsync_futex_waitv` should be completely removed

Tk-Glitch:
Thanks for all your efforts Iglu47 :frog: :heart: :heart: :heart:

Issue #949, opened 2023-01-31
sharedgpures-fences.patch fails to apply on mainline wine


csutcliff:
Hi, recently (I think it was working last week) the shared gpu resources patch "sharedgpures-fences.patch" has failed to apply to mainline wine.

```
sharedgpures-fences.patch -- ( Applied shared fences support patchset )
patching file dlls/winevulkan/make_vulkan
Hunk #1 FAILED at 99.
Hunk #2 succeeded at 113 (offset 1 line).
Hunk #3 succeeded at 129 (offset 1 line).
Hunk #4 succeeded at 197 (offset 1 line).
Hunk #5 succeeded at 217 (offset 1 line).
Hunk #6 succeeded at 251 (offset -1 lines).
Hunk #7 succeeded at 261 (offset -1 lines).
Hunk #8 succeeded at 286 (offset -1 lines).
Hunk #9 succeeded at 311 (offset -1 lines).
Hunk #10 succeeded at 1144 (offset -1 lines).
Hunk #11 succeeded at 1514 (offset -1 lines).
Hunk #12 succeeded at 1524 (offset -1 lines).
Hunk #13 succeeded at 1929 (offset -1 lines).
1 out of 13 hunks FAILED -- saving rejects to file dlls/winevulkan/make_vulkan.rej
patching file dlls/winevulkan/vulkan.c
patching file dlls/winevulkan/vulkan_loader.h
patching file dlls/winevulkan/vulkan_private.h
```

```
cat src/wine-git/dlls/winevulkan/make_vulkan.rej
--- dlls/winevulkan/make_vulkan
+++ dlls/winevulkan/make_vulkan
@@ -99,10 +99,8 @@ UNSUPPORTED_EXTENSIONS = [
     "VK_EXT_full_screen_exclusive",
     "VK_GOOGLE_display_timing",
     "VK_KHR_external_fence_win32",
-    "VK_KHR_external_semaphore_win32",
     # Relates to external_semaphore and needs type conversions in bitflags.
     "VK_KHR_shared_presentable_image", # Needs WSI work.
-    "VK_KHR_win32_keyed_mutex",
     "VK_NV_external_memory_rdma", # Needs shared resources work.

     # Extensions for other platforms
```

llde:
will take a look later, seems trivial

Tk-Glitch:
[...]

PR #950, opened 2023-02-01
Rebase shared gpu resource for wine git

llde:
Fix patch application in #949


Tk-Glitch:
Thanks @llde ! :frog: :heart:

Issue #979, opened 2023-04-01
"ERROR: Patch application has failed" on hotfixes/ow2/1148-staging.mypatch


GlitchedCode:
Trying to build wine-tkg on Arch Linux yields the following error.
```
Applying hotfix wine-tkg-patches/hotfixes/ow2/1148-staging.mypatch
patching file dlls/ntdll/tests/file.c
Hunk #1 succeeded at 4166 (offset 195 lines).
Hunk #2 FAILED at 5500.
1 out of 2 hunks FAILED -- saving rejects to file dlls/ntdll/tests/file.c.rej
patching file dlls/ntdll/unix/file.c
Hunk #1 FAILED at 1268.
Hunk #2 succeeded at 1769 (offset 220 lines).
Hunk #3 succeeded at 1799 (offset 232 lines).
Hunk #4 FAILED at 1580.
Hunk #5 FAILED at 1598.
Hunk #6 FAILED at 1617.
Hunk #7 FAILED at 1662.
Hunk #8 FAILED at 2232.
Hunk #9 FAILED at 3946.
Hunk #10 FAILED at 3988.
Hunk #11 FAILED at 4015.
Hunk #12 FAILED at 4193.
Hunk #13 FAILED at 4224.
Hunk #14 succeeded at 5435 (offset 1099 lines).
Hunk #15 succeeded at 5471 (offset 1099 lines).
Hunk #16 succeeded at 5481 (offset 1099 lines).
Hunk #17 succeeded at 5489 (offset 1099 lines).
Hunk #18 succeeded at 5508 (offset 1099 lines).
Hunk #19 succeeded at 5600 (offset 1099 lines).
Hunk #20 succeeded at 5614 (offset 1099 lines).
Hunk #21 succeeded at 5656 (offset 1112 lines).
Hunk #22 FAILED at 4589.
Hunk #23 FAILED at 4600.
Hunk #24 succeeded at 5927 (offset 1135 lines).
Hunk #25 succeeded at 5985 (offset 1177 lines).
13 out of 25 hunks FAILED -- saving rejects to file dlls/ntdll/unix/file.c.rej
```

Not sure what causes this, but removing the `wine-tkg-git/wine-tkg-git/wine-tkg-patches/hotfixes/ow2` folder seems to momentarily circumvent the issue.

rchris494:
guess i did not search very well. i see the problem is already reported.

PR #985, opened 2023-04-05
Disable 1148 hotfix on staging be92c5d+

therealmate:
fixes https://github.com/Frogging-Family/wine-tkg-git/issues/979 (i think)


Tk-Glitch:
Can't thank you enough! 🐸 ❤️ ❤️ ❤️

PR #1002, opened 2023-05-05
Update proton-tkg meta patchsets for bd89ab3 and move previous versions to legacy

Nanotwerp:


Nanotwerp:
Actually woahhhhh nevermind! patch applies but build fails. the issue is a little more complex than I thought :broken_heart: . don't want to mess anything up so I'm closing for now!

PR #1004, opened 2023-05-06
Fix proton 8 on distros using /usr/lib

Bill5122:
This change was also needed on gentoo which uses /usr/lib instead of /usr/lib32.

Untested on other distros using /usr/lib.


Fox2Code:
I think adding this flag unconditionally is a bad idea.

You should make a folder check like `[ ! -d "/usr/lib/glib-2.0/include" ]` before adding the flag.

By the way, thank for doing the PR and helping improve wine-tkg!

Tk-Glitch:
It's similar to what we did on other distro checks so we can give that a go as is. We'll add more checks if needed on both fronts.

Tk-Glitch:
Thanks Bill5122 ! :frog: :heart:

PR #1010, opened 2023-05-21
Fix tkg starting scripts

Tuupertunut:
There were three issues in the starting scripts `bin/wine-tkg`, `bin/wine64-tkg`, `bin/wine-tkg-interactive` that are packaged with a complete wine-tkg.

- The `PATH` and `LD_LIBRARY_PATH` variables were not exported. That means the variables were not passed down to the wine binary and therefore were not doing anything.
- The ascii frog image printed at startup was printed to stdout. This caused the frog to be part of every script output that used wine, so wine-tkg could not be used in scripts. This issue was solved by printing the frog to stderr instead.
- There were no quotes around the `${@:1}` parameter given to wine. This caused any quotes around arguments to be stripped, so arguments with spaces would not work. Example: `wine-tkg cmd /c echo "a & echo b"` should print `a & echo b` but instead it was printing a and b on different lines.

### DIFF
--- a/wine-tkg-git/wine-tkg-scripts/wine-tkg
+++ b/wine-tkg-git/wine-tkg-scripts/wine-tkg
@@ -42,7 +42,7 @@ cat >&2 << 'EOM'
 EOM
 
 if [ -x "$(command -v _wine)" ]; then
-  _wine ${@:1}
+  _wine "${@:1}"
 else
-  wine ${@:1}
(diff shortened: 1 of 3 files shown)


Tk-Glitch:
I'm not sure those are still needed, but it definitely won't hurt to have them fixed.
Thanks Tuupertunut! :frog: :heart:

PR #1021, opened 2023-06-06
Disable syscall emu patchset on WoW64 builds

TheComputerGuy96:
It's currently broken in this case because it causes an error when running a 32-bit application in the WoW64 mode :frog:

Note: I targeted a high enough commit because I doubt WoW64 works on Wine v7.20 well enough to be usable

### DIFF
--- a/wine-tkg-git/wine-tkg-patches/hotfixes/earlyhotfixer
+++ b/wine-tkg-git/wine-tkg-patches/hotfixes/earlyhotfixer
@@ -275,6 +275,13 @@ EOM
       fi
     fi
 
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
Thanks TheComputerGuy96! :frog: :heart:

gabriele2000:
@TheComputerGuy96 Is the error something like `shellex: failed to execute` or something similar?

TheComputerGuy96:
@gabriele2000 I think it is 🐸

gabriele2000:
> @gabriele2000 I think it is frog

NICE, THANKS A LOT!
I was thinking that it was the new SysWOW64 implementation that was faulty, but that was pretty weird...

PR #1022, opened 2023-06-06
Fix ccache not being set on newer Wine versions/WoW64

TheComputerGuy96:
Wine commit b1f59bc679a8c2dea18a6789a5b9b1a1ae825129 made some changes to how CC variables work, so we have to set it for the individual architectures now

This also sets the i386 variable for 64-bit-only WoW64 builds (this is because of the way new-style WoW64 builds work)


Tk-Glitch:
Looks good to me. Thanks TheComputerGuy96 :frog: :heart:

PR #1023, opened 2023-06-06
Add a mainline version of the Proton child window patch

TheComputerGuy96:
There's only one (1) conflict when applying the staging version, so this fixed version should probably work fine :frog:


Tk-Glitch:
Sure enough :D
Thanks again TheComputerGuy96! :frog: :heart:

PR #1026, opened 2023-06-12
Fix chaotic-staging wine-tkg profile

dr460nf1r3:
2 of the patches are causing the build to fail currently, therefore they need to be removed. Fixes https://github.com/chaotic-aur/packages/issues/2300 🐸


Tk-Glitch:
Indeed. Thanks dr460nf1r3 ! :frog: :heart:

PR #1034, opened 2023-06-29
Hotfixes: Update Proton Bcrypt patches for RDR2

Riesi:
Pulled from Proton-8.0 tree.
I hope I haven't messed anything up!

I choose wine-8.10  `68a3b0077e64d1b5232ff75996b82766bcc64ced` as the commit, which seems to work with the new and the old versions of the patch. So it should be good :frog: 

The unconventional inverted ancestor check is hopefully ok. :grimacing: 
Pretty cool that we don't need any reverts here anymore!


Tk-Glitch:
Looks fine at first glance. The reverse check doesn't follow our usual style but that's fine. Thanks Riesi :frog: :heart:

PR #1036, opened 2023-07-02
Remove ntdll: stack guard patchset

llde:
After wine commit 871f333d55c2c472014a861bfdd256c1f497dd40 
Commit 871f333d
authored 2 months ago by Alexandre Julliard
ntdll: Use a separate memory allocation for the kernel stack.

The  ntdll: stack guard patchset started causing strange behaviours rendering entire applications unusable: some of them are
MTG:Arena installer (cause  address violation exception)
MSBUILD (error during compilation of  complex programs and crashes)
DOTNET (prefix with dotnet may fail to update becouse threads are prone  to crashing while holding locks, depending on the build settings and patchsets included)
Visual Studio 2019 (with this patchset it hang while with the proper patches can now work semi-properly)
Analysis of the changes of the 871f333d55c2c472014a861bfdd256c1f497dd40 commit seems to shows that the patch is unable to properly work at all, as the strategy of the thread stack allocation was deeply changed in the reported commit, but the code is similar enough in the zone affected by the patch that the patch remains valid and is applied.


Saancreed:
I think it would be a good idea to also remove those patches from legacy patchsets in `mainline/legacy/proton-tkg-$revision.patch` and `staging/legacy/proton-tkg-staging-$revision.patch` that are applied to older versions of Wine because without doing this, new builds of e.g. Wine 8.9 will also be broken.

llde:
I thought about it, and while I agree, isn't immediate to see which revision are affected.

Tk-Glitch:
Thanks llde ! I'll check which versions are affected on my end.
:frog: :heart:

PR #1038, opened 2023-07-04
Use jq instead of grep when downloading DXVK from GitHub

hwsmm:
It fixes `_use_dxvk="release"` in `proton-tkg.cfg`, tested on Ubuntu 22.04


Tk-Glitch:
Thanks hwsmm ! :frog: :heart:

PR #1039, opened 2023-07-06
Don't disable ntdll-Syscall-Emulation from 'abbfbb8e'

gabriele2000:
Starting from 'abbfbb8e' this patch is disabled by default, so trying to disable it again would trigger a build failure because the staging patch script wouldn't find the patch, hence the failure.


TheComputerGuy96:
@gabriele2000 Maybe I should've done the WoW64 check based on the staging patchset enablement status (and not the commit hash) but that would only allow quite old Wine builds to use WoW64 (which was quite useless back then)

Also I have 2 other things to share:
1) This check should be changed to only succeed with the new wow64 option for NOLIB32
2) I have a local workaround that allows WoW64 to work without disabling the syscall emulation patchset by disabling the seccomp stuff if WoW64 mode is detected at runtime

PR #1043, opened 2023-07-13
Disable LAA patch for WoW64 builds

ventureoo:
Large Address Aware seems to be not adapted to work with WoW64, so it causes an error during runtime: 
`wine client error:1a8: sendmsg: Bad file descriptor`
So I suggest disabling it for all WoW64 builds until it is fixed (probably?).


Tk-Glitch:
Thanks ventureoo! :frog: :heart:

PR #1045, opened 2023-07-18
Add wine64 symlink for WoW64 builds

ventureoo:
This fixes installation scripts such as winetricks and setup_dxvk for WoW64 builds that don't have the wine64 binary in them. Of course, this should be fixed on the side of the scripts themselves, but for now I suggest create symlink to make life easier for users.


Tk-Glitch:
Yeah that will do the trick for now. Thanks ventureoo ! :frog: :heart:

PR #1051, opened 2023-08-12
Use relative path instead of absolute path

IroAlexis:
Indeed when you want to move the construction or share it, the link is no longer correct.

Before:
```
$ ls -la .
-rwxr-xr-x 1 iroalexis wheel  16488 Aug 12 14:21 wine
lrwxrwxrwx 1 iroalexis wheel    110 Aug 12 14:23 wine64 -> /home/iroalexis/projects/wine-tkg-git/wine-tkg-git/wine-tkg-clean-mainline-git-8.13.r343.gb2a099b3cee/bin/wine
```
Now
```
$ ls -la .
-rwxr-xr-x 1 iroalexis wheel  16488 Aug 12 14:00 wine
lrwxrwxrwx 1 iroalexis wheel      4 Aug 12 14:02 wine64 -> wine
```


Tk-Glitch:
Indeed! Thanks IroAlexis :frog: :heart:

PR #1052, opened 2023-08-12
Adjust wine64 symlink fix for use it too into PKGBUILD

IroAlexis:
I forgot the PKGBUILD build even if it's less serious but it's cleaner and adopts the same behavior as Wine's tools (winecpp for example) :frog:


Tk-Glitch:
Thanks again IroAlexis ! :frog: :heart:

PR #1067, opened 2023-09-16
Hotfixes: Fix Proton Bcrypt patches for RDR2

Riesi:
Since https://github.com/wine-mirror/wine/commit/2fd33f6ce10f67cfe553f6bac9a8a853c997f9ab
The patches were broken again.
I had to work around an assert for wow64, but the patches never worked for it anyways so it should not be an issue.

Also good news for the future of these patches! The missing gnutls parts have been merged upstream, so wine can implement the current proton patches properly soon.
https://gitlab.com/gnutls/gnutls/-/issues/894

Should be in `GnuTLS 3.8.2 (Aug 5, 2023–Oct 15, 2023) `, so only a few more months with these hacky patches. :)


Tk-Glitch:
Thanks Riesi! Also great news going forward :)
:frog: :heart:

PR #1106, opened 2023-12-15
Removed "libtiff-devel.i686" for fedora to reduce package conflicts

AvoMC:
`libtiff-devel.i686` caused no installation of the x86 dependencies. Removing this package from the dependency installation should fix the package conflict and continue to build `wine-tkg-git`


Tk-Glitch:
Thanks AvoMC! :frog: :heart:

PR #1138, opened 2024-03-14
[IMPROVE] Better deps resolver

loopyd:
**PR Tier:**  **Feature**

# Issue

``_ci_build`` is intended to offer unattended build functionality for ci pipelines.  It currently doesn't as the package manager stops and prompts the user for confirmation.

The dependency resolver is just spaghetti in ``deps`` and lacking distro-agnostic functionality.  It really should handle a lot more edge cases, especially with priviledge escalation and package management options. Because we have a lot of users on different system configurations.   We live in a nix world where everyone has preferences so our code has to react dynamically to that.  It did not before this PR.

# What's changed?

- The dependency resolver now continues without prompting the user if they set the ``_ci_build`` parameter to ``"true"``, offering the intended functionality of an unattended installation.  It is at this point that the user has acknowledged that installation should continue without providing them warnings, and they know what they are doing.

- The ``-y`` flag is inserted for apt on debuntu systems, and dnf on OpenSUSE-based systems.  pacman based distros get the ``y`` flag added onto their command line.

- A slight modification passes ``_ci_build`` as a function parameter to avoid circularly sourcing the ``prepare.sh`` script.

- A heavy cleanup has been done.  I tried to leave these changes little.  But they where indentation issues and ``shellcheck`` warnings I felt needed a skilled touch. 
- ``_nonmakepkg_dependency_resolve_distro`` is now a boolean ``"true"`` / ``"false"`` flag named ``_nonmakepkg_dependency_autoresolver`` because OS autodetection is done by improved logic in rhe ``deps`` script. 

- The ``deps`` script gains dynamic function calling where package dependencies per distro and arch is managed by simple bash arrays.  Making it easily maintainable. A single public wrapping functions ``install_deps`` and ``remove_deps`` wraps everything. 

- Auto making decisions for package manager and priviledge escalation options in ``deps``.  Supports ``apt``, ``dnf``, ``rpm``, ``pacman`` and for priviledge escalation ``sudo``, ``gosu`` and ``doas``

- Integration and fix up of logic in ``build.sh`` and ``prepare.sh`` for dependency resolution done right. 

- Noticed and fixed the fact that you only get warned to fully understand what options you are enabling if you are in a debuntu distro running ``nonmakepkg_build.sh`` interactively.  Now this warning appears properly for everyone.  Because not only debuntu users are guilty of skipping the steps to read and understand responsibility of setting ``_ci_build`` to ``"true"``.  ;)

- You may notice that I restored ``customization.cfg`` back to its original self from a symlink, I'd done this in testing when running this project in a docker container.  Nothing has actually changed, just placing this back so the PR runs smoothly.

- Added mirror autoresolution (currently only works in debuntu systems) and retry functionality that helps improve dependency installation success for packages like ``docbook-utils`` and ``libgstreamer0-dev:i386`` on debuntu systems.

# Goals of this PR

I just want working clean code for autoresolution of build dependencies that will not destroy system state and do things cleanly on every distro the project supports. 

I want to make it easier for other developers to contribute support for building wine in other distros by making the code more modular and maintainable. 

# My lillypad

I plan to also support Nala for debuntu based distros where that functionality forks from choosing between apt and Nala.  Because Nala is just better at dependency resolution and has rollback options where apt doesn't.  That however is a bit out of scope so I am leaving that out of this PR

# Ribbit

If you like this PR, don't like it or have feedback on room for improvement leave that for me.  I wanna hear those croaks.


wine-tkg-git/wine-tkg-scripts/build.sh
```diff
@@ -121,38 +121,37 @@ _build_serial() {
...
     # nomakepkg
-    if [ "$_nomakepkg_midbuild_prompt" = "true" ]; then
+    if [[ "$_nomakepkg_midbuild_prompt" == "true" ]]; then
       msg2 '64-bit side has been built, 32-bit will follow.'
       msg2 'This is the time to install the 32-bit devel packages you might need.'
       read -rp "    When ready, press enter to continue.."
     fi
-    if [ "$_nomakepkg_dep_resolution_distro" = "debuntu" ]; then
-      _debuntu_32
+    if [[ "$_nomakepkg_dependency_autoresolver" == "true" ]]; then
```
Tk-Glitch, line 139: We probably want to keep the style consistent and mimic the less bash'd line 146

wine-tkg-git/customization.cfg
Tk-Glitch, line 1: Where's our .cfg gone :fearful: ?

loopyd: Restored below with changed doc line for ``_nomakepkg_dependency_autoresolver``

wine-tkg-git/wine-tkg-scripts/build.sh
loopyd: Resolved with [2c0b5e5](https://github.com/Frogging-Family/wine-tkg-git/pull/1138/commits/2c0b5e5bd364e3b90cc93a39636d95ce114b1868) and https://github.com/Frogging-Family/wine-tkg-git/pull/1138/commits/df0f5a7e5250add6be91e42ca0398df5d0409fd8

loopyd:
**Updates:**

# Auto-Retry functionality

``_install_pkg`` and ``_remove_pkg`` now retry up to 3 times, triggering some additional functionality on the repeat attempts to attempt to resolve the problem.

# Mirror Resolution

Inclusion of mirror autoresolution with ``apt-smart`` on debuntu systems.  The code contains forks for the other OS's to perform automatic update of the mirrors, but no functionality is implemented for those distros yet.  Mirror update is triggered during the second and third retry installation of a dependency to try to attempt to find a working mirror that contains the package that was missing or not found.

This actually got the code working again in jammy when canonical broke some things in jammy this morning (missing packages in security), and can help with that in the future by using mirrors to download packages.

# Critical exits

The build now aborts from ``_build.sh`` and from ``nonmakepkg_build.sh`` when if specified, properly when ``_nonmakepkg_dependency_autoresolver="true"``, the operation fails and prompts the user to check the output.

wine-tkg-git/wine-tkg-scripts/package-debian.sh
```diff
@@ -86,16 +86,17 @@ Depends: libc6 (>= 2.17), libfontconfig1 (>= 2.11), libfreetype6 (>= 2.2.1), lib
...
-tar czf $tmpDir/control.tar.gz -C $tmpDir/control ./ >> /dev/null
+tar czf $tmpDir/control.tar.gz -C $tmpDir/control ./ >>/dev/null
 rm -r $tmpDir/control
 
 # Create .deb file
 prev=$PWD
 cd $tmpDir
-ar -r $name debian-binary control.tar.gz data.tar.gz &> /dev/null
+ar -r $name debian-binary control.tar.gz data.tar.gz &
+>/dev/null
```
Tk-Glitch, line 96: Not sure about that return, but that one script is kinda problematic anyway

Tk-Glitch:
Let's go! Thanks loopyd ! :frog: :heart:

PR #1162, opened 2024-04-26
deps: Add endeavouros and cachyos as arch

Artewar67:
Simple fix nomakepkg dependency autoresolver for endeavouros and cachyos


Tk-Glitch:

PR #1164, opened 2024-05-01
Fix _NUKR="debug" for valve wine builds

peterjc123:


Tk-Glitch:
Thanks peterjc123! :frog: :heart:

PR #1165, opened 2024-05-03
ntsync5: Update all ntsync5 patch files

Artewar67:
Change ntsync5 non-staging version because is based on slighly older version
Also fix the staging version by adding the "server_protocol.h" file to the patch files


Artewar67:
Revert server-Do-not-signal-thread-until-it-is-really-gone for ntsync need for two reason
1: It's not necessary for ntsync
2: It causes a slight slowdown

Tk-Glitch:

PR #1166, opened 2024-05-07
Update customization.cfg and proton-tkg-cfg

Artewar67:
Remove obsolete information from customization.cfg about the incompatibility of ntsync5 with _use_staging and _protonify. Also add two options: _ntsync5 and _wayland_driver for proton-tkg.cfg


Tk-Glitch:
I'll have to do something about proton, but that's alright. Thanks Artewar67! :frog: :heart:

PR #1168, opened 2024-05-08
ci: bump workflow actions to latest version

SoulHarsh007:
This PR bumps the workflow actions to the latest version. This is done to ensure that the workflows are using the latest features. Older versions of the actions may not be supported in the future and may cause the workflows to fail.

I have compared the CI outputs on my end, and they are the same, some outputs even have better compression ratios for the artifact tarball. You can also verify it from actions results on my fork: https://github.com/SoulHarsh007/wine-tkg-git/actions/

### DIFF
--- a/.github/workflows/proton-arch-nopackage.yml
+++ b/.github/workflows/proton-arch-nopackage.yml
@@ -11,7 +11,7 @@ jobs:
     container: archlinux:latest
 
     steps:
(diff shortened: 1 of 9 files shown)


Tk-Glitch:
Nothing seems to break. Thanks SoulHarsh007! :frog: :heart:

PR #1178, opened 2024-05-17
Update proton-tkg patches and move old versions to legacy

Artewar67:
Update proton-tkg patches and move old ones to legacy due to incompatibility with this commit: https://gitlab.winehq.org/wine/wine/-/commit/a5e707f7a1ed7f65ef6484ffcdbf2ee597bad030


Tk-Glitch:
Thanks Artewar67 ! You beat me to it :frog: :heart:

PR #1180, opened 2024-05-21
Hotfixes: Add de-steamify patch for Proton-9.x

Thaodan:


.dir-locals.el
Tk-Glitch, line 1: Not sure we want emacs footprints in here :p

Thaodan: Just take it won't bother all the non Emacs users anyway :p
I tried to do this editor neutral impossible.

Tk-Glitch:
Merging

PR #1186, opened 2024-05-31
gstreamer.sh: Fix building cause by GCC 14 and FFmpeg

Artewar67:
Fixing a gstreamer build with two commits
1: Adding parameters for CFLAGS and CXXFLAGS for GCC 14
2: Using an older version of FFmpeg caused by removing a needed function: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/3378
https://github.com/FFmpeg/FFmpeg/commit/9ee59b63f5ea37700916815538c32dbcadc4a512


Tk-Glitch:

PR #1189, opened 2024-06-09
ntsync5: Update ntsync5 patches, fastsync script and move older version to legacy

Artewar67:


Tk-Glitch:
Thanks yet again Artewar67 :frog: :heart: :heart:

PR #1190, opened 2024-06-09
childwindow-proton: Update patches and move older version to legacy

Artewar67:


Tk-Glitch:
You beat me to it! Thanks Artewar67 :heart: :frog: :frog: :heart:

Issue #1191, opened 2024-06-11
ntsync5 patch failed


damachine:
helllo become this failed message: (last week compiling was succesfull)
any help what  did wrong?

`
Install Wine-tkg. . .
Klone nach 'wine-tkg-git'...
remote: Enumerating objects: 19337, done.
remote: Counting objects: 100% (3299/3299), done.
remote: Compressing objects: 100% (738/738), done.
remote: Total 19337 (delta 2753), reused 2860 (delta 2561), pack-reused 16038
Empfange Objekte: 100% (19337/19337), 14.47 MiB | 5.18 MiB/s, fertig.
Löse Unterschiede auf: 100% (14706/14706), fertig.
git version 2.45.2
------------------
Project: wine-tkg-git (1 branch, 42 tags)
HEAD: ccf50d0a (master, origin/master)
Version: 7.x+
Created: 4 years ago
Languages:                           
           ● BASH (48.5 %) ● Shell (32.3 %)
           ● Python (19.2 %) 
Authors: 93% Tk-Glitch 2221
          1% Dmitry Skvortsov 33
          0% Torge Matthies 7
Last change: 17 hours ago
Contributors: 62
URL: https://github.com/Frogging-Family/wine-tkg-git.git
Commits: 2379
Churn (18): …/childwindow/childwindow-proton 4
            …/childwindow/childwindow-proton.patch 4
            …/childwindow/childwindow-proton-mainline.patch 3
Lines of code: 11011
Size: 228.89 MiB (1684 files)
  -> 
  ->        .---.`               `.---.
  ->     `/syhhhyso-           -osyhhhys/`
  ->    .syNMdhNNhss/``.---.``/sshNNhdMNys.
  ->    +sdMh.`+MNsssssssssssssssNM+`.hMds+
  ->    :syNNdhNNhssssssssssssssshNNhdNNys:
  ->     /ssyhhhysssssssssssssssssyhhhyss/
  ->     .ossssssssssssssssssssssssssssso.
  ->    :sssssssssssssssssssssssssssssssss:
  ->   /sssssssssssssssssssssssssssssssssss/
  ->  :sssssssssssssoosssssssoosssssssssssss:
  ->  osssssssssssssoosssssssoossssssssssssso
  ->  osssssssssssyyyyhhhhhhhyyyyssssssssssso
  ->  /yyyyyyhhdmmmmNNNNNNNNNNNmmmmdhhyyyyyy/
[...]

PR #1192, opened 2024-06-12
ntsync5: Fix ntsync5-mainline and ntsync5-protonify

Artewar67:
Should solve that issue: https://github.com/Frogging-Family/wine-tkg-git/issues/1191


Tk-Glitch:
Thank you Artewar67! :frog: :heart: :heart: :frog:

PR #1193, opened 2024-06-13
ntsync5: fixup ntsync5 patches

LionHeartP:


Tk-Glitch:
Thank you LionHeartP ! :frog: :heart: :frog:

PR #1197, opened 2024-06-15
childwindow-proton: Rebase patches and move older version to legacy

Artewar67:
Rebase patches for compatibility with this commit: https://gitlab.winehq.org/wine/wine/-/commit/2a2637743b82031d0039698e116a0d1daf18dc1d


Tk-Glitch:
Thanks yet again Artewar67! :frog: :heart:

Issue #1195, opened 2024-06-15
Error in prepare while applying ntsync5-staging.patch


pppbb:
[prepare.log](https://github.com/user-attachments/files/15845370/prepare.log)

[wine-tkg.cfg.txt](https://github.com/user-attachments/files/15845099/wine-tkg.cfg.txt)

I get error : `patch: **** malformed patch at line 3873:  { 
==> BŁĄD: Patch application has failed. The error was logged to /home/piotr/Pobrane/tmp/wine-tkg-git/wine-tkg-git/prepare.log for your convenience.
==> BŁĄD: Wystąpił błąd w prepare().`

Tk-Glitch:
Should be fixed with https://github.com/Frogging-Family/wine-tkg-git/commit/d64b87e602f70fd3b413c18dcce3de557e8b9d68

LionHeartP:
> [prepare.log](https://github.com/user-attachments/files/15845872/prepare.log) Unfortunately compilation is still broken for me.

Can I see your `include/config.h.in.rej` ?

pppbb:
`Can I see your include/config.h.in.rej ?`

[config.h.in.rej.txt](https://github.com/user-attachments/files/15854895/config.h.in.rej.txt)

pppbb:
I can confirm that PR #1198 works for me.

Tk-Glitch:
Hopefully resolved for a while with https://github.com/Frogging-Family/wine-tkg-git/commit/3e43387665eebdeaae86c10b748ab92b0839c540

PR #1198, opened 2024-06-15
ntsync5: Fix patches

Artewar67:
Should fix that issue: https://github.com/Frogging-Family/wine-tkg-git/issues/1195


LionHeartP:
My `config.h.in` looks like this and therefore merging this commit will break it again for me and others that have already tested it.

```/* Define to 1 if you have the `posix_fallocate' function. */
#undef HAVE_POSIX_FALLOCATE

/* Define to 1 if you have the `prctl' function. */
#undef HAVE_PRCTL

/* Define to 1 if you have the `proc_pidinfo' function. */
#undef HAVE_PROC_PIDINFO
```

Artewar67:
The strange thing is that your patch is correct and matches the repository, but for some reason I have most of the "`" in config.h.in become '
Can you send customization.cfg and advanced-customization.cfg

Artewar67:
After doing a little research, I found that autoconf is modifying the files. But I don't know why yet

PR #1200, opened 2024-06-16
ntsync5: Fix apply patch when config.h.in is different from upstream

Artewar67:
For some reason autoconf is modifying several files, including config.h.in, which causes an error when applying the ntsync5 patch. I added two new patches, ntsync-config.h.in-alt and ntsync-config.h.in-org, which apply the changes depending on the line in config.h.in


Tk-Glitch:
Let's be honest, it's not the prettiest hack :D That will work for now for sure.
Thank you once more Artewar67 :frog: :heart:

PR #1203, opened 2024-06-17
hotfixes: Add new hotfix patch to fix winewayland.drv regression

Artewar67:
This patch should fix a problem with winewayland caused by changes in win32u

Also reverse old commit: https://github.com/Frogging-Family/wine-tkg-git/commit/014d66774d5884491b6328075e6cff0d3aa05d3d as it is no longer needed


Tk-Glitch:

Issue #1205, opened 2024-06-25
Build failure of wine-tkg (makepkg build)


SunilPaul174:
[prepare.log](https://github.com/user-attachments/files/15967119/prepare.log)
  -> Applying childwindow-proton.patch
==> ERROR: Patch application has failed. The error was logged to /home/pc/tkg/wine-tkg-git/wine-tkg-git/prepare.log for your convenience.
==> ERROR: A failure occurred in prepare().

SunilPaul174:
childwindow fix off:
https://pastebin.com/Ebvu90FH

PR #1206, opened 2024-06-25
childwindow-proton: rebase, move previous version to legacy, etc etc

whrvt:
ignore the brain fart fork name

fixes https://github.com/Frogging-Family/wine-tkg-git/issues/1205


Tk-Glitch:
Thanks whrvt! :frog: :heart:

PR #1230, opened 2024-08-05
Updated de-steamify-90.mypatch for currentent experimental

adolfintel:
Valve-exp-bleeding builds have been broken for about a week because a function in ntdll/loader.c has changed, this PR updates the de-steamify-90.mypatch file so it works on the current version.


Tk-Glitch:
Thank you adolfintel ! :frog: :heart: :frog:

PR #1235, opened 2024-08-09
childwindow-proton: Rebase patches, update script and move older version to legacy

Artewar67:


Tk-Glitch:
That's another thanks from all of us to add to your belt Artewar67 :heart: :frog: :heart:

Issue #1236, opened 2024-08-10
problem during building (Arch)


Sprusrroot:
error at the end of the log:
[prepare.log](https://github.com/user-attachments/files/16568996/prepare.log)
![wine_error](https://github.com/user-attachments/assets/0d41173f-9c00-48a7-92a7-0d9d3ac07332)

patching file dlls/winex11.drv/bitblt.c
Hunk #1 FAILED at 2263.
1 out of 1 hunk FAILED -- saving rejects to file dlls/winex11.drv/bitblt.c.rej

Sprusrroot:
Thanks a lot, it seems to have helped!
But now I have a new error, for which I could not find a solution, if it is not difficult, then help me
The lines of the error log and the log itself: 
1078 str: Hunk #2 FAILED at 166.
1088 str: 1 out of 11 hunks FAILED -- saving rejects to file dlls/ntdll/unix/system.c.rej
[prepare.log](https://github.com/user-attachments/files/16569368/prepare.log)

Sprusrroot:
Thanks, it worked!
But now there is a new problem at compile time: "NTSYNC_IOC_SEM_READ" is not described (the first use in this function); did you mean "NTSYNC_IOC_SEM_POST"?
[wine-tkg_compile_log_error.log](https://github.com/user-attachments/files/16570163/wine-tkg_compile_log_error.log)

Artewar67:
Can you show the prepare.log file?

Sprusrroot:
[Uploading prepare.log…]()

Sprusrroot:
Here:
[prepare.log](https://github.com/user-attachments/files/16570260/prepare.log)

Artewar67:
Did you also install `ntsync-dkms` and `ntsync-header` from AUR?

Sprusrroot:
Thank you, no. I'm going to try it now. I installed nsync from default repositories

Sprusrroot:
Thank you very much for your all help! Everything compiled fine

PR #1237, opened 2024-08-10
proton-cpu-topology-overrides: Rebase patch, script and move older version to legacy

Artewar67:
Should fix this: https://github.com/Frogging-Family/wine-tkg-git/issues/1236


Tk-Glitch:
Thanks Artewar67 :heart: :heart: :heart: :heart: :frog:

PR #1244, opened 2024-08-15
proton-tkg-addition: Rebase patch, rename older version and update script install

Artewar67:


Tk-Glitch:
Artewar67 to the rescue! Thanks you! :frog: :heart:

PR #1249, opened 2024-08-29
Remove unnecessary dependency (jxrlib)

therealmate:
Wine uses builtin jxrlib for a while now: https://gitlab.winehq.org/wine/wine/-/commit/cfc8f154e5a7a3231db5045d5eac129da1cb638e


Tk-Glitch:
Thank you therealmate! You're the real mate! :frog: :heart:

Issue #1255, opened 2024-09-10
[bug] battle.net launcher is broken on wine > 9.15


OneOfOne:
The patch from https://bugs.winehq.org/show_bug.cgi?id=29384#c56 fixes it, any chance this could be included?

Tk-Glitch:
Should work again! Thanks @Artewar67

PR #1256, opened 2024-09-10
add new hotfix: battlenet for fix Battle.Net segfaults

Artewar67:
After wine-staging 9.16 the ntdll-WRITECOPY does not contain 10 and 11 patch because of the problems they cause for tests: https://gitlab.winehq.org/wine/wine-staging/-/commit/c901884ac0af17eeeac309269693d959cfb45e0b. But they are necessary for battle.net and games running via EA desktop to work: https://github.com/Frogging-Family/wine-tkg-git/issues/1255


Tk-Glitch:
Nice catch! Thanks Artewar67 :heart:
:frog: :frog: :frog:

PR #1257, opened 2024-09-14
childwindow-proton: Skip patch

Artewar67:
Child window support has been merged into upstream so the patch is no longer needed: https://gitlab.winehq.org/wine/wine/-/merge_requests/6467


Tk-Glitch:
Yup, that's all good. Thanks Artewar67!
:frog: :heart:

PR #1262, opened 2024-09-20
battlenet-fix: Don`t apply patch on staging commit

Artewar67:
In wine-staging, patches to make Battle.net work have been returned: https://gitlab.winehq.org/wine/wine-staging/-/commit/e68e4dbb7529295dc4fde320689f80b1215a0419
The battle-fix patch will not be applied if we are on this commit

I've also redesigned the script to accept the patch only if staging patches are enabled and removed the _battle_fix variable, I think it's better to just apply the patch automatically, considering that it was only removed for two weeks


Tk-Glitch:
Yup that's better indeed. Thanks Artewar67! :frog: :heart: :frog:

PR #1263, opened 2024-09-20
opencl-fixup: Rebase patch

Artewar67:


Tk-Glitch:
We might want to nuke that patch down the line as the initial issue fixed by it isn't a thing anymore - at least on Arch.
Confirming it's okay on other distros as well will define if we can simply disable it altogether.

Thanks again Artewar67!

PR #1266, opened 2024-09-23
CI: Remove outdated steps from workflows

kangtastic:
As part of implementing better dependency autoresolution, a9d0a33afe ("[IMPROVE] Better deps resolver (#1138)") and ff45a43dd5 ("proton: Enable dependency autoresolver") removed the _nomakepkg_dep_resolution_distro option from customization.cfg and proton-tkg.cfg. However, CI workflows for non-makepkg builds of Wine and Proton still expect to set this option.

In addition, the workflow for Arch makepkg builds of Wine contains a workaround for Frogging-Family/wine-tkg-git#237 that was obviated in part by 37ce946749 ("fix pipewire-jack not providing jack2 anymore on arch linux anymore (#675)") and in part by 4971465b39 ("makepkg: Only grab lib32 deps when _NOLIB32 is set to "false").

Update the currently maintained CI workflows to remove these outdated steps. This shouldn't affect anything, as the steps being removed were already doing nothing.

### DIFF
--- a/.github/workflows/proton-arch-nopackage.yml
+++ b/.github/workflows/proton-arch-nopackage.yml
@@ -21,7 +21,6 @@ jobs:
           chown user -R /tmp
           export CARGO_HOME="$PWD"
           cd proton-tkg
-          sed -i 's/distro=""/distro="archlinux"/' proton-tkg.cfg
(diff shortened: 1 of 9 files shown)


Tk-Glitch:
Looks fine to me :)
Thank you kangtastic! :frog: :heart:

PR #1267, opened 2024-09-23
Add -ntsync in Wine build pkgname

kangtastic:
Also rework the logic for adding -esync and -fsync to make various conditions for adding these three pkgname components clearer. For example, ntsync can (theoretically) coexist with esync; in that case we would add just -ntsync, not -esync-ntsync (cf. esync and fsync coexistence).


Tk-Glitch:
I'm okay with that. Thanks kangtastic! :frog: :heart:

PR #1268, opened 2024-09-23
Clarify build requirements for NTSync5 in .cfg files

kangtastic:
ntsync is part of the kernel UAPI, so the presence of the ntsync module on the build host is not required to build Wine with ntsync support. A complete ntsync header file at /usr/include/linux/ntsync.h (replacing, if applicable, the incomplete stub mainlined in kernel 6.10) is sufficient.

Note also that the proposed ntsync patchset targets kernels >= 6.8, but builds still work if the build host is running an earlier kernel. This is most immediately relevant for CI, as the ubuntu-latest GitHub Actions runner has kernel 6.5.

### DIFF
--- a/proton-tkg/proton-tkg.cfg
+++ b/proton-tkg/proton-tkg.cfg
@@ -79,7 +79,9 @@ _use_GE_patches="true"
 _use_fastsync="false"
 
 # NTsync5 - https://repo.or.cz/wine/zf.git/shortlog/refs/heads/ntsync5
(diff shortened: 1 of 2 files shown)


Tk-Glitch:
Yup!
Thanks again kangtastic :frog: :heart:

kangtastic:
Thanks for accepting these small improvements.

From another Galaxy S kernel hobbyist, see you around :sweat_smile:

PR #1271, opened 2024-09-25
proton: Don`t use dxvk dxgi if using wined3d

Artewar67:


Tk-Glitch:
Nice catch Artewar67! Thanks :frog: :heart:

Issue #1265, opened 2024-09-23
Can't build wine-staging v9.18 with protonify="true"


rankynbass:
The build errors out with:
```
ccache gcc -m64 -c -o dlls/ntdll/unix/virtual.o ../wine-git/dlls/ntdll/unix/virtual.c -Idlls/ntdll -I../wine-git/dlls/ntdll \
  -Iinclude -I../wine-git/include -D__WINESRC__ -D_NTSYSTEM_ -D_ACRTIMP= -DWINBASEAPI= \
  -DWINE_UNIX_LIB -I/usr/include/x86_64-linux-gnu -Wall -pipe -fcf-protection=none \
  -fvisibility=hidden -fno-stack-protector -fno-strict-aliasing -Wdeclaration-after-statement \
  -Wempty-body -Wignored-qualifiers -Winit-self -Wlogical-op -Wno-packed-not-aligned -Wpointer-arith \
  -Wshift-overflow=2 -Wstrict-prototypes -Wtype-limits -Wunused-but-set-parameter -Wvla \
  -Wwrite-strings -Wno-misleading-indentation -fPIC -fasynchronous-unwind-tables -O2 -ftree-vectorize -Wno-error=implicit-function-declaration -Wno-error=incompatible-pointer-types -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=0
../wine-git/dlls/ntdll/unix/virtual.c: In function ‘get_prot_str’:
../wine-git/dlls/ntdll/unix/virtual.c:1127:52: error: ‘VPROT_WRITTEN’ undeclared (first use in this function); did you mean ‘VPROT_WRITE’?
 1127 |     buffer[3] = (prot & VPROT_WRITECOPY) ? (prot & VPROT_WRITTEN ? 'w' : 'W')
      |                                                    ^~~~~~~~~~~~~
      |                                                    VPROT_WRITE
../wine-git/dlls/ntdll/unix/virtual.c:1127:52: note: each undeclared identifier is reported only once for each function it appears in
../wine-git/dlls/ntdll/unix/virtual.c: In function ‘try_map_free_area’:
../wine-git/dlls/ntdll/unix/virtual.c:1304:11: warning: unused variable ‘ptr’ [-Wunused-variable]
 1304 |     void *ptr;
      |           ^~~
../wine-git/dlls/ntdll/unix/virtual.c: In function ‘NtProtectVirtualMemory’:
[...]

PR #1272, opened 2024-09-25
hotfixes: Add new hotfix legacy-ntdll-writecopy

Artewar67:
Apply this patch when protonify and staging is enabled

Should fix this: https://github.com/Frogging-Family/wine-tkg-git/issues/1265


Tk-Glitch:
I feel we're in for a ride pretty soon :sweat: 

PR #1279, opened 2024-09-30
hotfixes: Add new shm_esync_fsync hotfix to workaround high CPU usage

### LINKED ISSUE #1278: High CPU usage with esync/fsync enabled after Wine commit 54ca1ab6
Just open anything in Wine with esync/fsync enabled and check the CPU usage while moving the mouse around.
[Here](https://gitlab.winehq.org/wine/wine/-/merge_requests/5970/commits) is the MR which started causing this behavior.
[This patch](https://github.com/ValveSoftware/wine/commit/9ceae59cebd111f0fb06d63ad1495ea1becd714b) in Proton was made to work around this issue, which we should probably use until it's fixed upstream. I'll open a pull request shortly to add this as a hotfix.
# Before patching:

# After patching:

whrvt:
And enable it if esync/fsync is used on trees containing Wine commit [54ca1ab607d3ff22a1f57a9561430f64c75f0916](https://gitlab.winehq.org/wine/wine/-/commit/54ca1ab607d3ff22a1f57a9561430f64c75f0916).
The patch is just a rebase of [this](https://github.com/ValveSoftware/wine/commit/9ceae59cebd111f0fb06d63ad1495ea1becd714b), which was made to work around the same issue in Proton.

Somebody should make a bug report for wine-staging about this, since it affects non-tkg staging esync as well...

Fixes #1278

### DIFF
--- a/wine-tkg-git/wine-tkg-patches/hotfixes/hotfixer
+++ b/wine-tkg-git/wine-tkg-patches/hotfixes/hotfixer
@@ -45,6 +45,7 @@ if [ "$_unfrog" != "true" ]; then
   _hotfix="NosTale" _hotfix_boundary="" _hotfix_import
   _hotfix="battlenet" _hotfix_boundary="" _hotfix_import
   _hotfix="legacy_ntdll_writecopy" _hotfix_boundary="" _hotfix_import
+  _hotfix="shm_esync_fsync" _hotfix_boundary="" _hotfix_import
 
   # Custom hotfix patches with a commit boundary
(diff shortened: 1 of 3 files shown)


Tk-Glitch:
Thanks whrvt! :frog:
Elizabeth Figura is most likely already aware (esync/fsync creator) since Rémi made this, but this might be too much of a hack to be acceptable in staging. Proton and -tkg seem like ok places to use this, until ntsync delivers us, maybe.
:heart:

PR #1282, opened 2024-10-04
hotfixes: Don`t apply legacy-ntdll-WRITECOPY when WOW64 mode is enable

Artewar67:


Tk-Glitch:
Thank you Artewar76 yet again! :frog: :heart:

Issue #1202, opened 2024-06-16
Trying to use with UMU launcher


IdrisKalp:
Hi. Normally I use proton-tkg with steam and heroic without problems. I also tried it with umu launcher but I couldn't manage to run it. I got an error like:

`steamrt3 is up to date
Proton: Upgrading prefix from None to 9.0-999 (/home/idris/Games/ttp2-proton/)
wine: could not load ntdll.so: /home/idris/.local/share/Steam/compatibilitytools.d/proton_tkg_experimental.bleeding.edge.9.0.102702.20240614/files/lib/ntdll.so: cannot open shared object file: No such file or directory`

There was no problem with proton-ge, however.

This was the command I used:

`WINEPREFIX=/home/idris/Games/ttp2-proton GAMEID=the-talos-principle-2 PROTONPATH=/home/idris/.local/share/Steam/compatibilitytools.d/proton_tkg_experimental.bleeding.edge.9.0.102702.20240614/ mangohud umu-run /media/f/Games/TTP2/Talos2/Binaries/Win64/Talos2-Win64-Shipping.exe`

Artewar67:
in umu-launcher the developers have added the possibility to disable pressure vessel by exporting environment variable `UMU_NO_RUNTIME=pressure-vessel`, which is necessary for proton-tkg to work, but still needs some improvements from our side.

Wait, soon I will release merge request to fix it.

Artewar67:
https://github.com/Frogging-Family/wine-tkg-git/pull/1283

IdrisKalp:
Thanks for heads up and sorry for late answer. Is there any update on that commit above?

yvvki:
> Thanks for heads up and sorry for late answer. Is there any update on that commit above?

Dead, and it's currently not working even with `UMU_NO_RUNTIME=1` envvar.

PR #1283, opened 2024-10-09
Proton-TKG: Add experimental support for umu-launcher

Artewar67:
It adds the following things:

Modifies the proton startup script to run correctly with umu-launcher

Adds umu-protonfixes that are needed for the launcher to work.

Changes the version of the Steam runtime library to Sniper

Also another question: I want to change the version of Proton to 9, but there is a question. Are there any problems with this version?

### DIFF
--- a/proton-tkg/proton-tkg-profiles/advanced-customization.cfg
+++ b/proton-tkg/proton-tkg-profiles/advanced-customization.cfg
@@ -9,11 +9,11 @@
 # Some options will be missing from this config file compared to wine-tkg-git as they are enforced.
 
 # Proton branch to target for lsteamclient libs and steam helper on 4.x+ - When using a Wine 3.x base, "proton_3.16" branch will be enforced
-_proton_branch="experimental_8.0"
+_proton_branch="experimental_9.0"
 
 # Set to "true" to use the "noruntime" proton toolmanifest if available
 # Set to "sniper" to use the experimental sniper runtime
-_nosteamruntime="false"
+_nosteamruntime="sniper"
 
 # Proton SDL Joystick support, xinput hacks and other gamepad additions. _gamepad_additions depends on _sdl_joy_support.
 # This might be required for some FFB steering wheels, but can break gamepad support in some games.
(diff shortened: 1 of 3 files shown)


Artewar67:
To run with umu-launcher you also need to put the environment variable UMU_NO_RUNTIME="pressure-vessel" so that it would start.

loathingKernel:
> To run with umu-launcher you also need to put the environment variable `UMU_NO_RUNTIME="pressure-vessel"` so that it would start.

This env variable, disables the use of Sniper runtime in umu. If it is required for proton-tkg to work with umu, it is likely that it is not being built in the runtime correctly.

Artewar67:
> This env variable disables the use of Sniper runtime in umu. Umu presently consider this a debugging-only feature. If it is required for proton-tkg to work with umu, it is likely that it is not being built in the runtime correctly.

The problem is that wine, dxvk and vkd3d are still being built against the system libraries and compiler in our script, and this prevents umu-launcher from running it https://github.com/Frogging-Family/wine-tkg-git/issues/1202 https://github.com/Open-Wine-Components/umu-launcher/issues/99

Need to change the build script so that proton-tkg is built against steam runtime. In the meantime, umu launcher support is currently implemented rather poorly and experimentally.

loathingKernel:
That makes sense. What I find odd is setting the `required_tool_appid` in `toolmanifest.vdf` to Sniper. This will cause these builds to not work in Steam as Steam will force Sniper in this case. So I am a bit confused what was the intention here.

Artewar67:
I apologize for accidentally removing the branch.

Artewar67:
I used proton-tkg from the upstream wine branch and which games worked and which still need something else customized

Two games on Unity: No problems, everything started correctly.

Barony: No problems either

Pacific Drive (UE4): Game crashes because when running umu-launcher proton-tkg VKD3D_FEATURE_LEVEL was set to 12_1 but my video card doesn't support rtx and I had to set 12_0 myself to get the game to start up

Stalker Anomaly: The game needs to install d3dx* and d3dcompiler_* via winetricks to run.

Artewar67:
I also found one problem

My patch caused regression and in the file compatibiltytool.vdf the proton version is not written and the default TKGVERSION is left

Tk-Glitch:
Weird. Will check that out too. Thanks Artewar67!

beh-10257:
Damn

Artewar67:
I haven't worked on it in a while. It needs to be redone, I won't be doing it again for a while.

Tk-Glitch:
As a side note, we are now building Valve-based proton builds against the latest container by default, which fixes one of the blockers for this.
Ideally we'd want to be able to also build upstream-based builds that way (which would fix the proton-tooling meta-issue), but this requires additional work currently. This might get easier with proton 11 and updated SDK though.

PR #1289, opened 2024-10-19
Hotfixes: Update NosTale hotfix for 0bfdac7

kangtastic:
Fix broken builds since Wine upstream 0bfdac7f65
("win32u: Move the PE side DND callbacks to user32.").

Fixes Frogging-Family#1288.

PR #1290, opened 2024-10-19
NosTale: rebase patch and rename older patch

Artewar67:


Tk-Glitch:
Fixed with https://github.com/Frogging-Family/wine-tkg-git/pull/1289

PR #1292, opened 2024-10-22
prepare: Add NTsync version tag

kangtastic:
The Wine version tag with ntsync on is currently "TkG Plain" for mainline and "TkG Staging Esync" for staging builds.

"TkG Plain NTsync" and "TkG Staging NTsync" seem better.

Note that "Esync" shouldn't appear in the tag for staging ntsync builds. It happens because _use_staging="true" also forces _staging_esync="true", which then results in "Esync" being added to the tag even though the staging impl of esync is patched in only temporarily.


Tk-Glitch:
That's correct. Thanks kangtastic! :frog: :heart:

PR #1293, opened 2024-10-22
ntsync5: Rebase for wine-staging cd2cce2

kangtastic:
Rebase ntsync5 patches broken by wine-staging cd2cce28cc ("eventfd_synchronization: Fix compile warnings.").

Also revert the existing hotfix stopgap ecd65dc53a ("staging_fixes: Revert esync patchset for ntsync5 (#1242)").

PR #1297, opened 2024-10-24
ntsync5: rebase patches for cebad42f

Artewar67:


wine-tkg-git/wine-tkg-patches/misc/fastsync/ntsync5-staging-protonify.patch
```diff
@@ -831,29 +831,29 @@ index edfeb03..0000000
 -
```
kangtastic, line 831: Hi, thanks for this and your past contributions.

The next few lines here revert changes in #1293 and breaks builds on staging again.

wine-tkg-git/wine-tkg-patches/misc/fastsync/ntsync5-staging.patch
```diff
@@ -831,29 +831,29 @@ index edfeb03..0000000
 -
```
kangtastic, line 831: Same here, breaks builds on staging.

l12436:
There are seems missing some extra modify in mainline cause wineserver crash after patch
This is how I fixed.
Do not if correct, but it seems working

[fix-fastsync.diff.gz](https://github.com/user-attachments/files/17515043/fix-fastsync.diff.gz)

Artewar67:
With staging I did not have this commit https://github.com/Frogging-Family/wine-tkg-git/pull/1293 and therefore these patches were written without it

While I'm dealing with Steam and its crashes, I don't recommend merging this request for now.

kangtastic:
Yes, looking over the most recent Wine MR that broke our ntsync5 patch:

https://gitlab.winehq.org/wine/wine/-/merge_requests/6630/

Enough has changed that it might not be so simple this time.

Artewar67:
Looks like I was able to fix the steam. @l12436 Can you test it with this patch?
[0001-fix-ERROR_INVALID_HANDLE-for-steam-with-ntsync-patch.mypatch.zip](https://github.com/user-attachments/files/17524307/0001-fix-ERROR_INVALID_HANDLE-for-steam-with-ntsync-patch.mypatch.zip)

Artewar67:
Now the error of applying a patch for staging no longer appears and a fix for steam has been added

Tk-Glitch:
Thanks Artewar67 and kangtastic!
:frog: :heart: :frog: :heart:

l12436:
The patch that merged is missing some modify I added. three more no_fast_sync for those who use `struct object_ops`.
That will cause wineserver crash

PR #1298, opened 2024-10-26
ntsync5: fix crash wineserver

Artewar67:
Adds no_get_fast_sync in object.c which I only added in the protonify-staging patch


Tk-Glitch:
:+1: 
Thank you Artewar67 !
:frog: :heart:

Issue #1296, opened 2024-10-24
ERROR: Patches application has failed


Kerrung:
[prepare.log](https://github.com/user-attachments/files/17508177/prepare.log)
[last_build_config.log](https://github.com/user-attachments/files/17508179/last_build_config.log)
[proton-tkg.log](https://github.com/user-attachments/files/17508181/proton-tkg.log)
![image](https://github.com/user-attachments/assets/9f88bc07-2a6c-49b9-98fe-da1054007263)

s0mebodyhelpme:
Getting a patch application, albeit a different one, too:
`==> WARNING: Skipping OPWR-proton on current childwindow patchset
  -> Applying LAA-unix-staging.patch
  -> Applying proton-tkg-staging-nofsync.patch
==> ERROR: Patch application has failed. The error was logged to /home/al/rep/wine-tkg-git/wine-tkg-git/prepare.log for your convenience.
==> ERROR: A failure occurred in prepare().`

[prepare.log](https://github.com/user-attachments/files/17515041/prepare.log)

s0mebodyhelpme:
Disabling the proton-tkg-staging patchset still yields the following with ntsync patchset:
```
  -> Applying ntsync5-staging.patch
==> ERROR: Patch application has failed. The error was logged to /home/al/rep/wine-tkg-git/wine-tkg-git/prepare.log for your convenience.
==> ERROR: A failure occurred in prepare().
    Aborting...
```
[prepare.log](https://github.com/user-attachments/files/17515060/prepare.log)

Tk-Glitch:
Should be fixed after https://github.com/Frogging-Family/wine-tkg-git/commit/26f451ba853181638f1d49d12ada0ff505aade62 and https://github.com/Frogging-Family/wine-tkg-git/commit/68f0226940abc84696e737f5a53783881ecfd48e

Kerrung:
[proton-tkg.log](https://github.com/user-attachments/files/17528201/proton-tkg.log)
[prepare.log](https://github.com/user-attachments/files/17528205/prepare.log)
[last_build_config.log](https://github.com/user-attachments/files/17528206/last_build_config.log)
68f0226940abc84696e737f5a53783881ecfd48e
![image](https://github.com/user-attachments/assets/ea2aed8d-2689-467e-beab-933b33b4af66)

Kerrung:
@Artewar67 @kangtastic @l12436

[...]

PR #1299, opened 2024-10-26
proton-tkg: Add forgoten patch for mainline

Artewar67:
When rebasing patches I forgot to add the patch for mailline
Should fix this #1296


Tk-Glitch:
I missed it too! Thanks Artewar67!!
:frog: :heart:

Issue #1300, opened 2024-10-27
wine-staging-9.20 fails to build with fsync patches applied


BGazotti:
Compilation fails with the following error:

`server/object.c:112:5: error: initialization of ‘unsigned int (*)(struct object *, enum fsync_type *)’ from incompatible pointer type ‘void (*)(struct object *, struct wait_queue_entry *)’ [-Wincompatible-pointer-types]
  112 |     no_satisfied,               /* satisfied */
      |     ^~~~~~~~~~~~
server/object.c:112:5: note: (near initialization for ‘apc_reserve_ops.get_fsync_idx’)`

Full build log: 
[wine-9.20.log](https://github.com/user-attachments/files/17534448/wine-9.20.log)

9.20 wine release, with staging patches and fsync-unix-staging + fsync_futex_waitv

Thank you!

Artewar67:
You need to add `-Wno-error=implicit-function-declaration -Wno-error=incompatible-pointer-types` to _GCC_FLASG and _CROSS_FLAGS

BGazotti:
Those are set by default by the wine build system and seem to be there for
a good reason. Implicit function declarations are usually someone
forgetting to do an #include.

Is there any other known way? This feels like stuff will break at runtime.

Thank you kindly!

BGazotti:
Update: the patch misses changes to server/object.c in 9.20. Adding these lines appears to fix it.

```
diff --git a/server/object.c b/server/object.c                                                                                    
@@ -109,7 +109,8 @@                                                                                                               
     NULL,                       /* remove_queue */                                                                               
     NULL,                       /* signaled */                                                                                   
     NULL,                       /* get_esync_fd */                                                                               
+    NULL,                       /* get_fsync_idx */                                                                              
     no_satisfied,               /* satisfied */                                                                                  
     no_signal,                  /* signal */                                                                                     
     no_get_fd,                  /* get_fd */                                                                                     
     default_map_access,         /* map_access */                                                                                 
@@ -134,7 +135,8 @@                                                                                                               
     NULL,                      /* remove_queue */                                                                                
     NULL,                      /* signaled */                                                                                    
     NULL,                      /* get_esync_fd */                                                                                
+    NULL,                      /* get_fsync_idx */                                                                               
     no_satisfied,              /* satisfied */                                                                                   
     no_signal,                 /* signal */                                                                                      
     no_get_fd,                 /* get_fd */                                                                                      
[...]

PR #1301, opened 2024-10-28
fsync: rebase for cebad42

Artewar67:
Should fix this: #1300


Tk-Glitch:
Thanks Artewar67 !!
:frog: :heart:

PR #1309, opened 2024-11-09
cleanup: Remove LoL fix patch

EliasTheGrandMasterOfMistakes:
LoL now uses vanguard anticheat that acts in kernel level which means that is impossible to running currently on Wine Is better to remove it.


Tk-Glitch:
Thanks EliasTheGrandMasterOfMistakes!
:frog: :heart:

Issue #1327, opened 2024-11-23
Can't login in Battlenet with Proton-tkg


Kerrung:
Hi! Everything works fine with the inbuilt Steam's vanilla Proton Experimental. But with TKG the LOGIN button never appears, instead all I see is an infinitely long time spinning Battlenet logo signifying loading. This behaviour happens if I use FSYNC. But no logo even rotates if I try to use NTSYNC, because then the Battlenet application always hangs completely.
[last_build_config.log](https://github.com/user-attachments/files/17879079/last_build_config.log)
[prepare.log](https://github.com/user-attachments/files/17879080/prepare.log)
[battlenet_log.txt](https://github.com/user-attachments/files/17879159/battlenet_log.txt)
![Screenshot From 2024-11-23 13-31-58](https://github.com/user-attachments/assets/389aa26b-c20e-403b-b823-e1b9c520bd57)

AllenTzul:
Maybe you can try applying this patch [proton_battleye.patch](https://raw.githubusercontent.com/SingleInfiniteLoop/wine-tkg-git/78cea96f16139c5bed24593899c4defc8bfc8383/wine-tkg-git/wine-tkg-patches/proton-tkg-specific/proton_battleye/proton_battleye.patch).

Artewar67:
Looks like ntdll-WRITECOPY patch stop working properly on newer version. I tested and this occurs regardless of the synchronization method selected, including server-side. I also tested standard wine and wine-staging from arch linux repository and same thing

Artewar67:
[wine-tkg-userpatches.zip](https://github.com/user-attachments/files/17882775/wine-tkg-userpatches.zip)
Try these patches. download them and extract them into the wine-tkg-userpatch folder. Also set _protonify=false and clean the prefix

Kerrung:
> [wine-tkg-userpatches.zip](https://github.com/user-attachments/files/17882775/wine-tkg-userpatches.zip) Try these patches. download them and extract them into the wine-tkg-userpatch folder. Also set _protonify=false and clean the prefix

@Artewar67 I have just tested with your patches and fresh prefix, but have not changes, it have not helped.

Artewar67:
I found out that you have to clean the prefix before testing. Now wine from arch linux repository no longer has this bug and I'm trying to figure out what wine-tkg patch is causing it

Artewar67:
[...]

Issue #1322, opened 2024-11-18
Baldur's Gate 3 have not launching with Proton-tkg anymore.


Kerrung:
I use Arch Linux, btw. All works fine with vanilla Proton Experimental, but with TKG 5a279bff0c796254d1670228581891569b774ece I see only "Running install script" for endless time. And also I have tried to delete prefix at all, but have not changes. And also Steam doesn't give any information in the terminal while launching the game. It seems like some kind of regressions because it had worked fine for me at the past with TKG with the same cfg.
[proton-tkg.log](https://github.com/user-attachments/files/17807241/proton-tkg.log)
[last_build_config.log](https://github.com/user-attachments/files/17807242/last_build_config.log)
[prepare.log](https://github.com/user-attachments/files/17807243/prepare.log)
![image](https://github.com/user-attachments/assets/82e60854-fc88-48b1-8a40-99f36085d3a7)

Artewar67:
Found out it was my mistake. when I rebased the patches I made the error ntsync5-staging.patch

Kerrung:
@Artewar67 I have tested  with 17b6e8af983b6ccda9040460badfd3565ade1a78, but it still have not worked, no changes.

Kerrung:
@Artewar67 now works fine with FSYNC or with NTSYNC too since cadda7ef00fc9db5b90d4cc376a151766421d46a. Thank you!

PR #1331, opened 2024-11-26
ntsync5: Fix mainline and protonify patch

Artewar67:
This patches doesn`t have code for completion.c which cause wineserver crash in Battle.Net

Should fix this #1327 and possibly this #1322


Tk-Glitch:
Nice catch! Thanks Artewar67 :frog: :heart:

PR #1332, opened 2024-11-28
Update fastsync

grgbnc:
correct ` _ntsync` to ` _use_ntsync` in error msg


grgbnc:
I was referring to the `_use_ntsync` option in wine-tkg-git's .cfg, not the linux-tkg's .cfg

Tk-Glitch:
But the error mentions the linux-tkg's option. The error implies you're using the wine-tkg option already as you wouldn't see it otherwise, so mentioning it would make zero sense here. Also the wine option has absolutely no effect on the missing header situation.

grgbnc:
> But the error mentions the linux-tkg's option

Oops, my bad, I overlooked it. My apologies.

PR #1333, opened 2024-11-28
Update customization.cfg

grgbnc:
Add clarification

### DIFF
--- a/wine-tkg-git/customization.cfg
+++ b/wine-tkg-git/customization.cfg
@@ -36,6 +36,7 @@ _GENERATE_DEBIAN_PACKAGE="false"
 
 # Override config with one of the presets from /wine-tkg-profiles dir - Leave empty to get a prompt (legacy profiles won't be listed)
 # Set to "none" to silence the prompt
+# Default tkg preset: "default-tkg",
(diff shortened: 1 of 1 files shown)


Tk-Glitch:
The "default-tkg" profile is empty. See https://github.com/Frogging-Family/wine-tkg-git/blob/master/wine-tkg-git/wine-tkg-profiles/wine-tkg-default-tkg.cfg and has the exact same effect as using "none".
It's there only to have an entry for the prompt but it's not actually a profile, it's the absence of one.
https://github.com/Frogging-Family/wine-tkg-git/blob/cadda7ef00fc9db5b90d4cc376a151766421d46a/wine-tkg-git/wine-tkg-scripts/prepare.sh#L352C1-L355C9

So maybe it's even more confusing to state that you can override config with that option then use an empty profile that will override nothing at all :sweat_smile:

Tk-Glitch:
Thanks for pointing out some unclear documentation.
But to be honest there are many parts of this project that are unclear. I'm pretty confident we don't have erroneous documentation, but the more power and options, the more ways to confuse the user. Making it simpler for the average user was also one of the goals of this project. But to do that without taking features away from advanced users or devs isn't that simple.
Feel free to share ideas with us to help making the experience more streamlined !

PR #1334, opened 2024-11-28
ntdll-WRITECOPY: add additional patch for mainline

Artewar67:
This patch missing is build without staging patches fix problem with login window for Battle.net #1327


Tk-Glitch:

Issue #1341, opened 2024-12-07
wine-tkg fails to build wine-10.0-rc1 when patching server/process.h


silverhikari:
wine-tkg-git version: latest master e9125cb

wine version: 10.0-rc1

preset selected: default-tkg(disabled staging)

issue: wine-tkg fails to build wine-10.0-rc1 when patching server/process.h

comparing the .orig and .reg it seems to be cause by the addition of `struct list          rawinput_entry;  /* entry in the rawinput process list */` when applying the esync-unix-mainline.patch

process.h.orig:

```
    const struct rawinput_device *rawinput_kbd;   /* rawinput keyboard device, if any */
    struct list          rawinput_entry;  /* entry in the rawinput process list */
    struct list          kernel_object;   /* list of kernel object pointers */
    struct pe_image_info image_info;      /* main exe image info */
};
```

process.h.rej:

```
     const struct rawinput_device *rawinput_kbd;   /* rawinput keyboard device, if any */
     struct list          kernel_object;   /* list of kernel object pointers */
     pe_image_info_t      image_info;      /* main exe image info */
+    int                  esync_fd;        /* esync file descriptor (signaled on exit) */
 };
```

Logs: https://paste.ee/p/btO1v

PR #1342, opened 2024-12-08
esync-mainline: rebase against wine 10.0-rc1

Artewar67:
Should fix this: #1341


Tk-Glitch:
Sorry for the delay. Thank you Artewar67.
:frog: :heart:

Issue #1343, opened 2024-12-19
Need rebase of NTSYNC and update to v7, please


Kerrung:
Everything have not working anymore with NTSYNC patches. I believe you have not needed any logs this time? And probably you want to know that NTSYNC v7 recently released. Thank you!

Kerrung:
Hi, @Artewar67 ! Thank you for your recent rebase. But it have not working for me.
[proton-tkg-build-crash.txt](https://github.com/user-attachments/files/18216540/proton-tkg-build-crash.txt)
[last_build_config.log](https://github.com/user-attachments/files/18216544/last_build_config.log)
[prepare.log](https://github.com/user-attachments/files/18216545/prepare.log)
![image](https://github.com/user-attachments/assets/96bde723-2678-431b-918e-3887be6ce024)

Artewar67:
You need to update the ntsync.h file in /usr/include/linux/
Question how did you install ntsync5. are you using a custom kernel?

Kerrung:
> You need to update the ntsync.h file in /usr/include/linux/ Question how did you install ntsync5. are you using a custom kernel?

@Artewar67 I have using [official Arch kernel](https://archlinux.org/packages/core/x86_64/linux/) with [ntsync-dkms](https://aur.archlinux.org/packages/ntsync-dkms), [ntsync-header](https://aur.archlinux.org/packages/ntsync-header), [ntsync-common ](https://aur.archlinux.org/packages/ntsync-common). I think the reason is this AUR packages [still using](https://aur.archlinux.org/packages/ntsync-dkms#comment-1002270) v5.

Artewar67:
I'm surprised that it doesn't work. The thing is that ntsync5-dkms from AUR uses v6, which is different from the v5 documentation, the code itself works identically to the v5 version

In this case. You can try again since in this pull request: #1345  I slightly rebased the old patches for v6

Kerrung:
> I'm surprised that it doesn't work. The thing is that ntsync5-dkms from AUR uses v6, which is different from the v5 documentation, the code itself works identically to the v5 version
> 
> In this case. You can try again since in this pull request: #1345 I slightly rebased the old patches for v6

I will try it, thank you!

Kerrung:
@Artewar67 I just tested 41f7e2477bb25179439698fa86613408ba9afed0. It now have compiled fine, but my games don't works.

Artewar67:
[...]

PR #1344, opened 2024-12-20
ntsync5: Rebase for ntsync5 v7 Kernel Driver

Artewar67:
For some reason CachyOS in kernel 6.12.6 update ntsync5 kernel driver to v7 which cause problem
Also fix this: #1343 

These patches are incompatible with past versions


Tk-Glitch:
Thanks Artewar67! :heart:
:frog: :frog: :frog:

PR #1345, opened 2024-12-21
ntsync5: Implement ntsync7 support and revert ntsync5 to v6 version

Artewar67:
So far at the moment ntsync5 v7 is only available in linux-cachyos and maybe in others, but ntsync5-dkms from AUR and linux-zen still use the v6 version. So I added ntsync7 patches and a new parameter for `_use_ntsync5`. ntsync5 patches are back to v6


Tk-Glitch:
I guess it's fine for now considering the current situation. Ntsync upstreaming needs to happen.

sfjuocekr:
This does not apply cleanly to mainline v9.22

PR #1348, opened 2024-12-26
ntsync5: Add legacy patch for wine v9.22 nonstaging

Artewar67:
This should solve the problem @sfjuocekr commented on in this pull request: #1345
>This does not apply cleanly to mainline v9.22


Tk-Glitch:

PR #1368, opened 2025-02-15
Added Mold linker for quicker builds.

AlleyPally:


Tk-Glitch:
Mold can't be used for PE afaik so that wouldn't work unless you build without MinGW, which is discouraged. There also seems to be some friction with some flags used by Wine that would lead to broken/unexpected behavior, which would be a problem. Correct me if I'm wrong though.

AlleyPally:
Little update: You are correct! I recently tried compiling wine32 and wine-stable from the AUR (Due to the WOW64 migration) and had the mold linker enabled. It compiled but always ran into segmentation faults whenever I tried running any of the wine commands. This was just for wine so imagine what would've happened with proton! After turning off mold, it compiled correctly and nothing was off.

Mold is really good but it still has some ways to go before it becomes a true drop-in replacement :(.

PR #1369, opened 2025-02-15
Fix wine 10.1, add fsync patches for older versions

### LINKED ISSUE #1363: wine 10.1 fails when applying fsync-unix-mainline
wine 10.1 fails when applying fsync-unix-mainline caused by sync.c in dlls/ntdll/unix failing to patch. this seems to be cause by the section it was editing being change located at 1700:

NEW:
```
if (alertable)
    {
        /* Since server_wait will result in an unconditional implicit yield,
           we never return STATUS_NO_YIELD_PERFORMED */
        if ((status = server_wait( NULL, 0, SELECT_INTERRUPTIBLE | SELECT_ALERTABLE, timeout )) == STATUS_TIMEOUT)
            status = STATUS_SUCCESS;
        return status;
    }
```

OLD:
```
if (alertable) return server_wait( NULL, 0, SELECT_INTERRUPTIBLE | SELECT_ALERTABLE, timeout );
```

logs: https://paste.ee/p/WkYqt734

HengiFettlich:
Fixes #1363


Tk-Glitch:
Thank you HengiFettlich! :frog: :heart:

Issue #1373, opened 2025-02-18
Stream can't find wine64 file because it name is wine64-tkg. But, why???


Kerrung:
[last_build_config.log](https://github.com/user-attachments/files/18836322/last_build_config.log)
[prepare.log](https://github.com/user-attachments/files/18836323/prepare.log)
[SC2-part1.log](https://github.com/user-attachments/files/18836324/SC2.log)
Hi! I use Arch Linux, btw. When I try to start StarCraft 2 using Steam with Proton-tkg, directly SC2 exec file, not using Battlenet, Steam by default have tried to use wine64. But Steam can't find wine64 file because its name is wine64-tkg.

`FileNotFoundError: [Errno 2] No such file or directory: '/usr/share/steam/compatibilitytools.d/proton_tkg_makepkg/files/bin/wine64'`

So, then I have created a simlink
`sudo ln -s /usr/share/steam/compatibilitytools.d/proton_tkg_makepkg/files/bin/wine64-tkg /usr/share/steam/compatibilitytools.d/proton_tkg_makepkg/files/bin/wine64`
And so, then now Steam can to find `/usr/share/steam/compatibilitytools.d/proton_tkg_makepkg/files/bin/wine64`, but the game still can't start because an another error message which is looped spaming endlessly:
[SC2-part2.log](https://github.com/user-attachments/files/18836865/SC2-part2.log)

Artewar67:
This looks like cause be this commit: https://gitlab.winehq.org/wine/wine/-/commit/8c3f205696571558a6fae42314370fbd7cc14a12
>loader: Install the Wine loader in the Unix lib directory.

>And get rid of the "wine64" naming, the loader is always named "wine"
on all platforms.

Kerrung:
> This looks like cause be this commit: https://gitlab.winehq.org/wine/wine/-/commit/8c3f205696571558a6fae42314370fbd7cc14a12
> 
> > loader: Install the Wine loader in the Unix lib directory.
> 
> > And get rid of the "wine64" naming, the loader is always named "wine"
> > on all platforms.

I tried to create simlink to `wine`, not only to `wine64`, but it haven't helpes to start the game:
`sudo ln -s /usr/share/steam/compatibilitytools.d/proton_tkg_makepkg/files/bin/wine  /usr/share/steam/compatibilitytools.d/proton_tkg_makepkg/files/bin/wine64`

Artewar67:
Can you check how this work now?

[...]

PR #1379, opened 2025-02-21
PKGBUILD, non-makepkg-build, scripts: Better version of fix makefiles

### LINKED ISSUE #1374: makedep: error: mkdir  : No such file or directory
config.status: executing Makefile commands
makedep: error: mkdir  : No such file or directory
config.status: error: could not create Makefile
==> ОШИБКА: Произошел сбой в build().

[last_build_config.log](https://github.com/user-attachments/files/18872110/last_build_config.log)

[prepare.log](https://github.com/user-attachments/files/18872114/prepare.log)

Artewar67:
This check commit one time and should work with proton-tkg

Should fix #1374 and possibly #1373

### DIFF
--- a/proton-tkg/proton_template/conf/proton
+++ b/proton-tkg/proton_template/conf/proton
@@ -423,7 +423,7 @@ class Proton:
         self.wine_bin = self.bin_dir + "wine"
         # winetricks
         os.environ["WINE"] = self.bin_dir + "wine"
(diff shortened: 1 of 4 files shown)


whrvt:
this is okay, but FWIW, the single lib directory would also work for older wine versions, so you could just always used the new style without checking for the commit

PR #1385, opened 2025-02-22
build.sh: Add _new_makefiles check in _package_makepkg().

logos128:
Fixes https://github.com/Frogging-Family/wine-tkg-git/issues/1374 for external install with makepkg builds.


Artewar67:
Thank @logos128

PR #1407, opened 2025-03-15
dxvk_config: Disable patch on wine 7.02 and higher

Artewar67:
This patch no longer using in Proton since 7.0 so i think me can disable this


Tk-Glitch:
This is wrong. Setting `_dxvk_dxgi="false"` will enable the undesirable path of using wine's dxgi combined with dxvk. This will break many games.
I reverted your commit and decided to simply get rid of the whole dxvk_config support https://github.com/Frogging-Family/wine-tkg-git/commit/a7e04182d9a16333fddd826f651421339883cc9c

PR #1410, opened 2025-03-17
Proton-TKG NTSync build action

Syntist:
It would be nice if we can have ntsync build, although I know we need NTSync patch needed to be available in kernel, we can probably update the readme, so people can know what is required for NTSync to work?


Syntist:
Sorry, It's having issue building, so am fixing it right now.

Tk-Glitch:
Let's give it a whirl.
Thanks Syntist! :frog: :heart:

PR #1431, opened 2025-04-03
Properly name the NTSYNC action

XenHat:
The ntsync action was using the same name as the regular one, making it difficult to single out from the rest


Artewar67:
Thank XenHat

PR #1454, opened 2025-04-30
fsync_futex_waitv: fix patch

noisecode3:
This made it build again. It's just setting a new line number for `AC_CHECK_HEADERS`. Should it be a legacy `fsync_futex_waitv_configure_ac.patch`? Should it be split up? I could add that just point me in the right direction or I could look up how it was done.

This is [the one](https://github.com/wine-mirror/wine/commit/013ec6e652ca1bc94b4bdefd8c577d15857bef68) that's problematic from last week


Tk-Glitch:
Sorry I have made the wanted changes so we can merge it as is. That should tell you exactly the usual approach for the repo though.
Thank you very much noisecode3 !!! :frog: :frog: :frog: :heart: :heart: :heart:

PR #1483, opened 2025-05-21
esync-mainline: rebase against to commits

Artewar67:


silverhikari:
still get the failled patching, it seems to still be trying to use the older patch, even though i have both git pull at the base of the git repo and "rm -rv src/" to make sure it downloaded the latest wine-10.8 git


https://paste.ee/p/InQ6brSR

PR #1556, opened 2025-07-06
Updated de-steamify-10.0-be.mypatch for current bleeding-edge

adolfintel:
Some minor things changed in wine.inf.in and caused the patch to fail to apply.


Artewar67:
LGTM. I need merge some preparation in #1559 like fix build with staging patches and separate de-steamify-10.0-be.mypatch between experimental and bleeding edge.

Thank @adolfintel

PR #1573, opened 2025-07-13
ADD FSR4 amdcxffx64.dll

HengiFettlich:
Add option to pull latest (or specified) amdcxffx64.dll for FSR4 upgrade. The patches are already present, so we might as well add it.
This also adds the option PROTON_FSR4_UPGRADE akin to proton-ge and enables it by default. 
I have tested it with OptiScaler and Cyberpunk 2077. OptiScaler picked up FSR4 without manually installing the DLL into the game directory.

### DIFF
--- a/proton-tkg/proton-tkg.cfg
+++ b/proton-tkg/proton-tkg.cfg
@@ -55,6 +55,13 @@ _use_dxvk="git"
 _proton_dxvk_hud=""
 _proton_dxvk_configfile=""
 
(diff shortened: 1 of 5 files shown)


Tk-Glitch:
Yeah, even if we disable it by default and in CI it's too easy for packagers to distribute it and point the finger at us. The proton script sounds like a better place indeed. Or alternatively, winetricks.

PR #1686, opened 2026-01-21
container fixes: Pass the correct flag to configure.sh if SELinux is enabled

ProjectSynchro:
This is required on systems with SELinux since containers won't be able to access any bindmounts and the Proton container engine test will fail.

This shouldn't cause problems on other distros (or ones where SELinux isn't enabled) since selinuxenabled will return a non-zero value if it is disabled (put present) or the command -v call will fail if it isn't found on PATH.

If you'd like me to adjust this in any way, let me know. :smile:


ProjectSynchro:
No prob!

PR #1688, opened 2026-01-28
Add fsync-git replacement to PKGBUILD

silentnoodlemaster:
allows pacman to replace the previous default package with the new default package


Tk-Glitch:
First, this would overwrite the previous value making this invalid code wise as is.
Secondly, there's no wine-fsync package, be it on AUR or main Arch repos, so the replace has nothing to replace.
And third, the wine-faudio package is no more, so the current replace could be nuked as it doesn't make sense anymore.

PR #1708, opened 2026-07-01
Rebase opencl hotfix against 018943e

RedSQL:
Rebases `opencl-fixup` hotfix to apply on newest Wine.

This patch previously did not apply when compiling 11.12 so I rebased it to work again. Everything compiles and works fine on my end now.

I hope I modified the hotfixes script correctly to account for older versions 🐸


Tk-Glitch:
Thanks RedSQL.
Looks all good! 🐸 ❤️

